PR #2, opened 2018-09-13
Feature/layout media query initial breakpoints

carlosvillu:


components/layout/mediaQuery/README.md
```diff
@@ -79,4 +79,21 @@ const LayoutMediaQueryBootstrap = LayoutMediaQueryFactory(BREAKPOINTS)
 </LayoutMediaQueryBootstrap>
 ```
 
-> **Find full description and more examples in the [demo page](https://sui-components.now.sh/workbench/layout/mediaQuery/demo).**
\ No newline at end of file
+### initialMediaQueries props
+
+This component accept and special prop call `initialMediaQueries`. This props is specially useful in SSR, because setting up you can force a media queries object, for environments where the `componentDidMount` never is called, like in Node.
```
miduga: `This component accepts a special prop called initialMediaQueries. This prop is specially useful in SSR in order to force a initial media query object that will be used in environments where the componentDidMount is never called, like in Node.`

components/layout/mediaQuery/src/index.js
```diff
@@ -57,10 +70,19 @@ const LayoutMediaQueryFactory = BREAKPOINTS =>
       if (!shallowEqual(result, params)) this.setState({params: result})
     }
 
+    shouldComponentUpdate(_, nextState) {
```
miduga: Why not using PureComponent instead?

carlosvillu: I tried, we had the rerender anyway.

PR #6, opened 2018-09-20
feat(dropdown/basic): allow handlers on items, and close list if clic…

ruben-martin-lozano:
In this PR we allow to:
- Item links to have a click handler
- Close list if document click is triggered
- Close list if item link click is triggered

To keep the current behaviour in all projects, default values to new features is set to false.

### DIFF
--- a/components/dropdown/basic/src/index.js
+++ b/components/dropdown/basic/src/index.js
@@ -27,14 +27,15 @@ class DropdownBasic extends Component {
   /**
    * Toggle menu state: expanded/collapsed.
    */
-  _toggleMenu = () => {
-    const {expanded} = this.state
+  _toggleMenu = expand => {
+    const expanded = expand || !this.state.expanded
 
-    this.setState({expanded: !expanded})
+    this._handleListener(expanded)
+    this.setState({expanded})
   }
 
   /**
-   * Mouser over event handler.
+   * Mouse over event handler.
    */
   _onMouseOver = () => {
     this.setState({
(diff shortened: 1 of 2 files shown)


components/dropdown/basic/src/index.js
```diff
@@ -70,14 +71,18 @@ class DropdownBasic extends Component {
...
-  _renderLink = ({text, url, target}, index) => {
+  _renderLink = ({onClick, target, text, url}, index) => {
     const Link = this.props.linkFactory
 
     return (
       <li key={index} className="sui-DropdownBasicMenu-listItem">
         <Link
           href={url}
           className="sui-DropdownBasicMenu-listLink"
+          onClick={e => {
```
carlosvillu: Don't create inline arrow functions to handle events. They result in more memory consuming.

components/dropdown/basic/src/index.js
```diff
@@ -118,7 +151,10 @@ class DropdownBasic extends Component {
             <ArrowButtonIcon svgClass="sui-DropdownBasic-buttonIcon" />
           </button>
         </div>
-        <div className="sui-DropdownBasicMenu">
+        <div
+          className="sui-DropdownBasicMenu"
+          ref={node => (this.wrapper = node)}
```
carlosvillu: We should start to use the new Ref API more align with the next React version.

components/dropdown/basic/src/index.js
ruben-martin-lozano: 👍

ruben-martin-lozano:
@carlosvillu feedback aplicado ;)

PR #7, opened 2018-10-01
feat(dropdown/basic): allow custom classname on button

ruben-martin-lozano:
Allow custom button className

### DIFF
--- a/components/dropdown/basic/src/index.js
+++ b/components/dropdown/basic/src/index.js
@@ -134,13 +134,17 @@ class DropdownBasic extends Component {
     const wrapperClassName = cx('sui-DropdownBasic', {
       'is-expanded': expanded
     })
+    const buttonWrapperClassName = cx('sui-DropdownBasic-buttonWrap', {
+      [button.className]: !!button.className
+    })
+
     return (
       <div
         className={wrapperClassName}
         onMouseOver={expandOnMouseOver ? this._onMouseOver : this._doNothing}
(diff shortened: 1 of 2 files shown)


components/dropdown/basic/src/index.js
```diff
@@ -134,13 +134,17 @@ class DropdownBasic extends Component {
     const wrapperClassName = cx('sui-DropdownBasic', {
       'is-expanded': expanded
     })
+    const buttonWrapperClassName = cx('sui-DropdownBasic-buttonWrap', {
+      [button.className]: !!button.className
```
carlosvillu: Classname brake the encapsulation of this component.

You must use an enum prop and apply the inner class properly depend on the enum value.

kikoruiz: Just for info... why do you need a custom classname?

miduga: He needs a custom className to just use a `re-Select` class in order to give to the dropdown the look'n'feel of Fotocasa. This is not ideal, as the re-Select itself should be a component instead just classnames to be used.

kikoruiz: As you say, it's not the ideal, and as far as I know, we've been taken a different approach for this specific cases by overriding the classnames from our sites containers... So, I don't know if it's a door we should open, aren't you happy by overriding classnames?

ruben-martin-lozano:
Not necesary at this moment.

PR #10, opened 2018-10-24
fix(ad/blockerExtensionDetector): deleted remove method

ddmarin94:
Detele "remove method" isn't supported in IE11


miduga:
But then you're not removing the div at all? You could do a conditional rendering in the render method to remove it.

PR #11, opened 2018-10-30
Feature/highlight dropdown items

JoanClaret:
We are aiming to remove a hardcoded CTA link, and display a highlight link in the dropdown user. 
To do so, we need to:

- Add highlight prop validation in `topbar user`
- Remove hardcoded CTA in `topbar user` (we'll add it in the domain config)
- Get highlight prop in `dropdown user`
- Style highlight items in `dropdown user`

Related JIRA: https://jira.scmspain.com/browse/MTR-15351
Related PR: https://github.schibsted.io/scmspain/frontend-mt--lib-motor/pull/361

### DIFF
--- a/components/dropdown/user/src/index.js
+++ b/components/dropdown/user/src/index.js
@@ -32,12 +32,18 @@ class DropdownUser extends Component {
     })
   }
 
-  _renderLink = ({text, url, icon: Icon, notifications}, index) => {
+  _renderLink = ({text, url, icon: Icon, notifications, highlight}, index) => {
     const Link = this.props.linkFactory
+    const linkClassName = cx('sui-DropdownUserMenu-listLink', {
+      'sui-DropdownUserMenu-listLinkHighlight': highlight
+    })
+    const iconClassName = cx('sui-DropdownUserMenu-listIcon', {
+      'sui-DropdownUserMenu-listIconHighlight': highlight
+    })
     return (
       <li key={`${text}-${index}`} className="sui-DropdownUserMenu-listItem">
-        <Link href={url} className="sui-DropdownUserMenu-listLink" title={text}>
-          <Icon svgClass="sui-DropdownUserMenu-listIcon" />
+        <Link href={url} className={linkClassName} title={text}>
+          <Icon svgClass={iconClassName} />
           <span className="sui-DropdownUserMenu-listText">{text}</span>
           {!!notifications && (
             <span className="sui-DropdownUserMenu-listNotification">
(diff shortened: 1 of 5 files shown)


components/dropdown/user/src/index.scss
```diff
@@ -123,7 +123,8 @@
         display: flex;
         position: relative;
 
-        &:hover {
+        &:hover,
+        &Highlight {
```
tureey: This is OK but I prefer to separate the `selectors`  to the `pseudo-classes`. And to do not duplicate code, you can always usa partials with the `%`.

Something like:
```
// This is not going to be generated as css
%hoverEffect {
    color: color: $c-accent;
}

&:hover {
   @extend %hoverEffect;
   // only hover styles here
}

&:Highlight {
   @extend %hoverEffect;
   // only highlight styles here
}
```

JoanClaret: Placeholders are fine, i like them. 
They'll output exactly the same code as my original variation but, to be honest, in this case I don't see any improvement because it's is adding just a color property `color: $c-accent;` but you are right, for other cases with more CSS declarations would be useful.

JoanClaret:
Merging because we are only 2 FE at motor today.

components/dropdown/user/src/index.scss
tureey: Like I said, a totally personal opinion! Well done btw :smile:

kikoruiz: @tureey, we haven't had good experiences with placeholders here in Schibsted, so somehow they are not allowed 😛. There are "safer" ways to approach it.

components/dropdown/user/src/index.js
```diff
@@ -32,12 +32,18 @@ class DropdownUser extends Component {
     })
   }
 
-  _renderLink = ({text, url, icon: Icon, notifications}, index) => {
+  _renderLink = ({text, url, icon: Icon, notifications, highlight}, index) => {
     const Link = this.props.linkFactory
+    const linkClassName = cx('sui-DropdownUserMenu-listLink', {
+      'sui-DropdownUserMenu-listLinkHighlight': highlight
```
kikoruiz: I know it's already merged, but you know me, and I like to review everything 😜. In this case, you should have use a state class, not a modifier. So that it should be something like this:
```
const linkClassName = cx('sui-DropdownUserMenu-listLink', {
       'is-highlighted': highlight
})
```

components/dropdown/user/src/index.scss
tureey: 👍  I am curious though about these experiences hahaha

PR #13, opened 2018-11-21
fix(topbar/composable): close dropdown only if click is outside the w…

ruben-martin-lozano:
…rapper

### DIFF
--- a/components/topbar/composable/src/sui-topbar-composable-dropdown/index.js
+++ b/components/topbar/composable/src/sui-topbar-composable-dropdown/index.js
@@ -22,7 +22,7 @@ export default class DropdownMenu extends Component {
   }
 
   closeMenu = ({target}) => {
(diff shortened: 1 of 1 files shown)


components/topbar/composable/src/sui-topbar-composable-dropdown/index.js
```diff
@@ -21,14 +21,10 @@ export default class DropdownMenu extends Component {
     )
   }
 
-  closeMenu = e => {
-    if (e.target === this.headerDOMEl) {
-      return
-    }
+  closeMenu = ({target}) => {
+    var isClickOutsideDropdown = !this.wrapper.contains(target)
```
midudev: Don't use `var`.

ruben-martin-lozano: lol

PR #16, opened 2018-11-23
feat(setting/item): added setting item

ferransimon:


components/setting/item/README.md
```diff
@@ -0,0 +1,23 @@
+# SettingItem
```
jcger: we need to update the readme

PR #18, opened 2018-11-30
feat(map/basic): add propertyId on markers

AitorRodriguez990:
### Highlight POIS

- [x] Add propertyId on makers using `add` event  be able to find on DOM when we do _mouseover_ on cards.

### DIFF
--- a/components/map/basic/src/leaflet/marker-manager.js
+++ b/components/map/basic/src/leaflet/marker-manager.js
@@ -31,6 +31,10 @@ class MarkerManager {
 
   createMarker(item, deprecatedLabelNoPrice) {
     const events = [
+      {
+        eventName: 'add',
+        eventHandler: ({target}) => (target._icon.id = target.Id)
+      },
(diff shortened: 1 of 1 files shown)


midudev:
This logic shouldn't be here as it's coupled to what you need in the Real Estate card and this should be agnostic.

ruben-martin-lozano:
Hi @midudev :)

In this PR, the `id` that is being added to each POI, is the current marker ID: https://github.com/SUI-Components/schibsted-spain-components/blob/77107c0dd9e89bd9373fb4b175fa86f5b5617983/components/map/basic/src/leaflet/marker-manager.js#L58 so it's not strictly coupled to FC, but, I agree with you that the value used to the marker ID is the propertyId (this is coupled).

This component needs a refactor to be fully decoupled as @AitorRodriguez990 mentioned, but this PR does not increase the current FC coupling, so I would deal this refactor in another issue.

midudev:
Thanks for the explanation, I though the `Id` was the `propertyId` but as Ruben mentioned the problem is in another part of the code. :+1:

PR #19, opened 2018-12-03
Feature/callback params execution

zecafa:


components/form/pta/src/index.js
```diff
@@ -5,19 +5,37 @@ import {paramsToQueryString} from './querystring'
...
+  componentWillMount() {
+    window.addEventListener(MESSAGE_EVENT_TYPE, this.handleMessage)
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener(MESSAGE_EVENT_TYPE, this.handleMessage)
+  }
+
+  handleMessage({data: {type}}) {
+    const {onSubmit, onError} = this.props
```
midudev: You might do these `required` props or having a defaultValue to avoid possible problems.

zecafa: These are optional function callback fields. Which default value should i set?

midudev: Default value then should be an empty function in order your code to not break if the user of the component doesn't pass a function as a prop.

components/form/pta/src/index.js
```diff
@@ -5,19 +5,36 @@ import {paramsToQueryString} from './querystring'
...
   constructor(props) {
     super(props)
 
-    const {formUrl: BASE_URL, ...settings} = this.props
+    const {formUrl: BASE_URL, onSubmit, onError, ...settings} = this.props
     const QUERY = paramsToQueryString(settings)
     const formUrl = `${BASE_URL}?${QUERY}`
 
+    this.handleMessage = ({data: {type}}) => {
+      type === SUBMIT_SUCCESS_EVENT_TYPE && onSubmit && onSubmit()
```
midudev: That could work as well :D I usually prefer to avoid this and just give a default value for the prop as an empty func, but this would do the work.

zecafa: 👍

PR #22, opened 2018-12-11
feat(map/basic): allow customise attribution text displayed at the bottom right of map.

frmedina:
This PR will allow passing by props the text we want to show at the bottom-right side of the leaflet map.
This property is called `attribution`, and you can see its documentation from leaflet in the following link:
https://leafletjs.com/reference-1.3.4.html#control-attribution

### DIFF
--- a/components/map/basic/src/leaflet/layer-manager.js
+++ b/components/map/basic/src/leaflet/layer-manager.js
@@ -7,7 +7,15 @@ export default class LayerManager {
     this.layers = {}
   }
 
-  createMapLayers({mapViewModes, maxZoom, minZoom, appId, appCode, id}) {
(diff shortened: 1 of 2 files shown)


components/map/basic/src/leaflet/layer-manager.js
```diff
@@ -18,8 +26,7 @@ export default class LayerManager {
         {
           app_code: appCode,
           app_id: appId,
-          attribution:
-            'Map &copy; 1987-2017 <a href="http://developer.here.com">HERE</a>',
+          attribution: attribution,
```
midudev: In this case, you could use the shorthand for assigning keys with the same value name.

PR #23, opened 2018-12-12
checkboxesWithParent new component

AitorRodriguez990:
### checkboxesWithParent new component

- [x] Create new component to show a list of checkboxes with parent checkbox.
- [x] Add example for new component.

#### Context

- Parent will be checked when all children are.
- When a child is not checked, the parent neither.
- On check parent all children will be checked (or unchecked, depending on parent status). 

Does anyone have any advice to improve this component? Thank you!


components/form/checkboxesWithParent/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@schibstedspain/sui-form-checkboxes-with-parent",
```
danivm: if this is not a sui component the name should be `@schibstedspain/react-form-checkboxes-with-parent`

AitorRodriguez990: Other components on this repository has same name convention. For example:
- "name": "@schibstedspain/sui-form-checkbox"
- "name": "@schibstedspain/sui-card-basic"

What to do?

AitorRodriguez990:
Context added.

components/form/checkboxesWithParent/package.json
danivm: these components were previously sui and after moved to this repository. all the new components in this repo sould be `@schibstedspain/react-xxxxx` like:

https://github.com/SUI-Components/schibsted-spain-components/blob/master/components/cmp/modal/package.json#L2

AitorRodriguez990: Package name updated, but we must take into account that all components on this repository has `sui-` prefix on their package names.

components/form/checkboxesWithParent/README.md
```diff
@@ -0,0 +1,23 @@
+# FormCheckboxesWithParent
+
+> Description
```
davecarter: Don't forget to add a little description to new components so well be enjoying such a great reading later 😃

components/form/checkboxesWithParent/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Usage
+
+### Basic usage
+```js
+import FormCheckboxesWithParent from '@schibstedspain/form-checkboxes-with-parent'
+
+return (<FormCheckboxesWithParent />)
+```
+
```
davecarter: If the component has styling is a good idea to add somewhere that you must to import styles to in `index.scss` file:
```css
// Add component styles to index.scss file:
@import '~@schibstedspain/form-checkboxes-with-parent/lib/index'
```

AitorRodriguez990: README updated with more information :)

Issue #26, opened 2018-12-19
Cmp component has no height in IE


tureey:
We've seen that the body of the CMP Component has no height in IE.
_See image attached_

![image 3](https://user-images.githubusercontent.com/16169890/50207534-07492c80-036f-11e9-8ded-b817902717fc.png)

PR #27, opened 2018-12-19
fix(cmp/modal): fix height of the innerMmodal in IE

tureey:
In the IE browser the body of the modal has no height. Giving a minimum height we ensure it's shown
correctly.

I decided `50 vh` just because in case the content of the `inner modal` does not height a lot, it will not look weird.

ISSUES CLOSED: #26

### DIFF
--- a/components/cmp/modal/src/index.scss
+++ b/components/cmp/modal/src/index.scss
@@ -151,6 +151,7 @@ $lh-sui-cmp-modal-title: $lh-xxl !default;
     font-size: $fz-sui-cmp-modal-inner;
     overflow-y: scroll;
     padding: $p-l;
+    min-height: 50vh; // fix for IE
   }
 
   &-footer {


components/cmp/modal/src/index.scss
```diff
@@ -151,6 +151,7 @@ $lh-sui-cmp-modal-title: $lh-xxl !default;
     font-size: $fz-sui-cmp-modal-inner;
     overflow-y: scroll;
     padding: $p-l;
+    min-height: 50vh; // fix for IE
```
kikoruiz: From my point of view, it's a really bad practice to make things for IE, but well, it's already merged :(

tureey: I would say the same with complicated supports but I think that the tradeoff is good.
It gives a lot of value with only 1 line IMO @kikoruiz

PR #36, opened 2019-01-23
Feature/add form select

midudev:
Add form/select component.

image


components/form/select/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@schibstedspain/sui-form-select",
```
danivm: `sui-form-select` or `react-form-select` ?

midudev: `@schibstedspain/react-form-select`

PR #38, opened 2019-02-05
Feature/perf dynamic rendering

midudev:
This component improves performance using [dynamic rendering technique](https://developers.google.com/search/docs/guides/dynamic-rendering) at Component level.

It uses the `userAgent` prop to detect if the user is a bot or not.

**If you're a bot**, no matter if you're on the server side or the browser side, the wrapped content will be always rendered.

 **If you're an user**:
  - On the server side, nothing is rendered.
  - On the client side, the content will be rendered when it is about to interesect the viewport.

You could completely disable the behaviour of the component using the prop `disabled` and thus it will render nothing until it is enabled. This could be useful to enable the component on the browser side when the user performs an action, like a click.


components/perf/dynamicRendering/src/withIntersectionObserver.js
```diff
@@ -0,0 +1,60 @@
...
+          {...this.props}
+          isVisible={isVisible}
+          innerRef={this.innerRef}
+        />
+      )
+    }
+  }
+}
+
+export default hocIntersectionObserverWithOptions()
```
carlosvillu: Have we call the function here ?!

jcger: there are two exports, we are using the other one

midudev: From the copy pasta. I can remove it. :)

PR #40, opened 2019-02-07
[perf/dynamicRendering] allow disable on server

carlosvillu:


components/perf/dynamicRendering/src/index.js
```diff
@@ -33,13 +33,17 @@ function checkUserAgentIsBot({userAgent}) {
...
+  force,
   height,
   userAgent
 }) {
   const isBot = checkUserAgentIsBot({userAgent})
   const isOnBrowser = typeof window !== 'undefined'
 
-  // if isBot, we return in server and client the content
+  // Force render in server and client
+  if (force) return children
```
kikoruiz: I don't understand the positive meaning of force if you don't want to make your children to render dynamically. Is not `avoid` or `skip` words more appropriate?

PR #44, opened 2019-02-20
fix(perf/dynamicRendering): fix problem with hydrating nulled component

midudev:
Render a div to be hydrated in the client as a null will not fire a hydratation and definitely won't be rendered.


components/perf/dynamicRendering/src/.tern-port
```diff
@@ -1 +0,0 @@
-63476
```
jcger: I guess this file should be gitignored

midudev: The thing is already in the `.gitignore` file. I'm removing it and I expect to not see it again. :D

PR #45, opened 2019-02-21
Perf/move to function

midudev:
Move list/link and link/basic from class to function.


components/link/basic/src/index.js
```diff
@@ -1,78 +1,70 @@
...
+  return (
+    <a
+      className={linkClassName}
+      href={url}
+      onClick={handleClick}
+      rel={rel}
+      target={target}
+      title={title}
+    >
+      {renderContent()}
```
carlosvillu: Inception!

PR #52, opened 2019-03-04
feat(list/bullet): fix error when displaying illustrations on responsive desktop device

AgonisticKatai:
fix error when displaying illustrations on responsive desktop with margin right

### DIFF
--- a/components/list/bullet/src/index.scss
+++ b/components/list/bullet/src/index.scss
@@ -30,6 +30,7 @@ $sz-list-bullet-illustration-l: 72px !default;
       #{$prefix}-illustration {
         height: $sz-list-bullet-illustration-l;
         margin-bottom: $m-xl;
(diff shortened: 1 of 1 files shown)


components/list/bullet/src/index.scss
```diff
@@ -46,6 +46,11 @@ $sz-list-bullet-illustration-l: 72px !default;
   }
 
   &-illustration {
+    .sui-ListBullet--responsive & {
+      @include media-breakpoint-up(l) {
```
JoanClaret: why not including it there?
https://github.com/SUI-Components/schibsted-spain-components/blob/master/components/list/bullet/src/index.scss#L32

JoanClaret: so you don't need the mq and the selector because they are already there

PR #53, opened 2019-03-04
feat(list/bullet): add smallFont prop

JoanClaret:
Adding smallFont prop so we can reuse this component in different pages.

### DIFF
--- a/components/list/bullet/src/index.js
+++ b/components/list/bullet/src/index.js
@@ -5,9 +5,10 @@ import cx from 'classnames'
 
 const BASE_CLASS = 'sui-ListBullet'
 
-const ListBullet = ({listItems, responsive}) => {
+const ListBullet = ({listItems, responsive, smallFont}) => {
   const listBulletClass = cx(BASE_CLASS, {
-    [`${BASE_CLASS}--responsive`]: responsive
+    [`${BASE_CLASS}--responsive`]: responsive,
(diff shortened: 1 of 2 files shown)


components/list/bullet/src/index.js
```diff
@@ -5,9 +5,10 @@ import cx from 'classnames'
 
 const BASE_CLASS = 'sui-ListBullet'
 
-const ListBullet = ({listItems, responsive}) => {
+const ListBullet = ({listItems, responsive, smallFont}) => {
   const listBulletClass = cx(BASE_CLASS, {
-    [`${BASE_CLASS}--responsive`]: responsive
+    [`${BASE_CLASS}--responsive`]: responsive,
```
rmoralp: Are two variants compatible?

JoanClaret: We does not have specification about it  :/

components/list/bullet/src/index.scss
```diff
@@ -30,12 +30,18 @@ $sz-list-bullet-illustration-l: 72px !default;
       #{$prefix}-illustration {
         height: $sz-list-bullet-illustration-l;
         margin-bottom: $m-xl;
-        margin-right: 0;
```
AgonisticKatai: We need this specification to work properly on desktop-responsive

JoanClaret: fixed. well seen mate.

PR #55, opened 2019-03-05
feat(card/arrow): link card with arrow icon

AgonisticKatai:
image

(**In order to be able to launch SUI-Studio locally, I had to fix some linter bugs in other components.**)


components/card/arrow/src/index.scss
```diff
@@ -0,0 +1,51 @@
+/* sass-lint:disable no-important*/
+@import '~@schibstedspain/sui-theme/lib/settings-compat-v7/index';
```
JoanClaret: Not necessary, i think you can delete it.

components/card/arrow/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+   */
+  icon: PropTypes.func,
+  /**
+   * Factory for the component that will hold the card link.
+   */
+  linkFactory: PropTypes.func,
+  /**
+   * URL for the link that wraps the whole card.
+   */
+  link: PropTypes.string
```
JoanClaret: `.isRequired`?

components/card/arrow/README.md
```diff
@@ -0,0 +1,23 @@
...
+$ npm install @schibstedspain/sui-card-arrow --save
+```
+
+## Usage
+
+### Basic usage
+```js
+import CardArrow from '@schibstedspain/sui-card-arrow'
+
+return (<CardArrow />)
```
JoanClaret: maybe we could explain how to use it with props

```
const link = 'https://www.lorem.com/'

const media = {
  alt: 'Lorem image',
  src: 'https://www.lorem.com/ipsum.jpg'
}

const text = {
  title: 'Lorem',
  description: 'Lorem ipsum dolor sit amet'
}

<CardArrow media={media} text={text} link={link}/>
```

components/card/arrow/src/index.scss
AgonisticKatai: In order to use the @include sui-card, I need @import '~@schibstedspain/sui-theme/lib/settings-compat-v7/index';

JoanClaret: are you sure? 
this mixin isn't located in the `settings-compat` https://github.com/SUI-Components/sui-theme/blob/master/src/components/_card.scss#L6

AgonisticKatai: Some of the variables used by sui-card are defined in settings-compat-v7. 🤷‍♂️

JoanClaret: ok, i see. This means that non deprecated components are using a deprecated config file.

Same happens with buttons
https://github.com/SUI-Components/sui-theme/blob/master/src/components/_button.scss#L1
Which is using a deprecated file
https://github.com/SUI-Components/sui-theme/blob/master/src/settings-compat-v7/_components.scss#L81

and some other components.

is this something that we expect @jcger ?
should we move this vars outside the `compat` folder?

anyways, this is not related with this PR, IMHO we can manage this in a further PR.

jcger: Yep, it's tech debt we will face once we end developing all of our atoms and molecules

components/form/pta/src/index.js
```diff
@@ -16,11 +16,17 @@ class FormPta extends Component {
   constructor(props) {
     super(props)
 
-    const {formUrl: BASE_URL, onSubmit, onError, eventListeners, ...settings} = this.props
+    const {
```
jcger: this changes shouldn't be part of this PR, right?

jcger: In this case it shouldn't be a big deal but you are changing form/pta stuff under the card/arrow scope so no release changes will be found

AgonisticKatai: ok, being very controlled changes, and not needing to make a new release of the form/pta, I upload it and in a future modification of the form/pta we solve it.
Thank you very much!

PR #85, opened 2019-06-18
feat(form/builder): refactor submit action

AgonisticKatai:
> Now, the component is no longer responsible for managing the submit of the form.
This is because for product needs, we are forced to add the login component and others with user data, between the form and the submit button.
For this reason, the management of errors, we must also perform in the wrapper component, to prevent the action of sending data if any field contains an error.

### DIFF
--- a/components/form/builder/src/index.js
+++ b/components/form/builder/src/index.js
@@ -384,6 +384,7 @@ FormBuilder.propTypes = {
   onInputChange: PropTypes.func.isRequired,
   /** Get the submit state of the form */
   isSubmited: PropTypes.bool,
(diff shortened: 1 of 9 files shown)


components/form/builder/src/index.js
```diff
@@ -189,15 +188,17 @@ class FormBuilder extends Component {
     return (
       <Select
         errors={errors}
-        showErrors={showErrors}
+        isSubmited={isSubmited}
```
kikoruiz: I think the correct spelling is `submitted` ;)

AgonisticKatai: ups... thx!

PR #87, opened 2019-06-19
feat(form/builder): update formbuilder show error management

AgonisticKatai:
> Change in the way errors are displayed. Now it depends on a prop that will indicate if it should show them or not.

### DIFF
--- a/components/form/builder/src/components/text-area.js
+++ b/components/form/builder/src/components/text-area.js
@@ -13,20 +13,17 @@ const TextArea = ({
   placeholder,
   size,
   onChange
-}) => {
-  // onError({[field]: errorText})
-  return (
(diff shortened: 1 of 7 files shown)


components/form/builder/src/components/text-area.js
```diff
@@ -7,16 +7,14 @@ import WithValidator from '../validatorHoC/WithValidator'
...
   id,
   value,
   placeholder,
   size,
-  onChange,
-  onError
+  onChange
 }) => {
-  onError({[field]: errorText})
+  // onError({[field]: errorText})
```
rmoralp: remove comments

AgonisticKatai: DONE!

PR #94, opened 2019-07-02
Feature/add abtesting context for experiments

desko27:
## ✍️ Description

The proposed additions to the AB Testing experience have been previously discussed between different members of Real Estate teams who are working every day in AB Testing and will definitely take advantage of these changes.

The following features are introduced in this PR:

- **Experiment Context**: OptimizelyXExperiment component provides a unique-per-experiment context with useful data like already-prepared variation booleans.
- **useExperiment**: A new `useExperiment` hook is added as a new package for experiment-related hooks. It returns the context of the experiment at any hierarchy level. Also, it returns a fallback object just in case the component where it's called from is not wrapped by any OptimizelyXExperiment.
- **forceVariation**: OptimizelyXExperiment accepts a new `forceVariation` prop which allows to force rendering a specific variation while ignoring defaultVariation or any optimizely response. This is ideal to test variations in local development scenarios.
- **variationName**: OptimizelyXExperiment children accept a new `variationName` prop to specify a name for each variation, which can be used later to identify the chosen variation from the context. If no variation name is passed to a specific child, it will take a letter from the alphabet which matches its position.

Despite all the new stuff, the basic functionality continues to work as before, so no breaking changes are introduced.

## ☑️ Tasks

- [x] Experiment-based context data for OptimizelyXExperiment children, new hook to consume it.
- [x] New props for OptimizelyXExperiment and its children.
- [x] Jest tests to ensure new props and context hook work as expected.
- [x] Docs for all the new features (see new "Advanced Usage" section).

## ⚠️ About Travis error

TL;DR This is not a problem to merge the PR.

New tests are failing on Travis due to an import of [something that is not yet published](https://github.com/SUI-Components/schibsted-spain-components/pull/94/files#diff-36bcc5f76adeea2ab4de868fb69d3cbbR2). When replacing the absolute import in local all the tests are working fine:
```js
import {ExperimentContext} from '../../../optimizelyXExperiment/src'
```
Screenshot 2019-07-03 at 09 41 20

But the import of the context must be kept absolute due to the way react context works. Tests will work in Travis once`@s-ui/abtesting-optimizely-x` package is published.


components/abtesting/optimizelyXExperiment/src/index.js
```diff
@@ -1,22 +1,119 @@
...
+    this.initialVariationId =
+      this._getForceVariationId() || this.defaultVariation
+    this.initialVariationName = this._getVariationNameFromVariationId(
+      this.initialVariationId
+    )
+
+    // no need to pass the special children prop to the context
+    const {children, ...propsWithoutChildren} = props
+
+    this.state = {
```
midudev: In order to simplify and avoid duplicating calculating some info, you could avoid using state for fields that you could calculate once on the `render` and pass on the object.

For example, `isDefault` and `isVariation` could be calculated on the render using the `variationId` from the state.

This will help you to simplify the code and do some checks only in one place.

desko27: I see, but I'm putting all that stuff in the state because of [this warning from the React docs](https://es.reactjs.org/docs/context.html#caveats), so the value is passed as a unique reference to the provider (the state itself) and not as a new object at every render.

Does it make any sense? Or am I missing something?

midudev: Make totally sense. What about extracting a function the creation of the state in order to be used for the initialState and the update? At least, you will reuse the state logic and keep it in one place.

desko27: Oh, *laik* 😄

desko27: ### ✅ Done

Tests keep passing in my local environment and it looks much better now, thanks!

components/abtesting/optimizelyXExperiment/README.md
```diff
@@ -41,8 +41,113 @@ Component will obtain OptimizelyX's decision for given `experimentId` and displa
...
 
-## Known issues
+## Advanced Usage
+
+### Experiment Context
+
+The following data is set in a plain object to the experiment context.
+
+- All props passed to OptimizelyXExperiment (see limitations at Known issues section).
+- isActive: true when a decision has been made by optimizely.
```
midudev: A suggestion to improve the README. You could use code styling for values that improves the readability of those.

```suggestion
- isActive: `true` when a decision has been made by optimizely.
```

desko27:
> That's an awesome work providing huge value to all developers creating experiments and using this component. Congrats dude 🥇

🙇🏻‍♂️ Thanks dude!

PR #97, opened 2019-07-08
feat(script/loader): load styles

JoanClaret:
Some scripts need to load styles in order to work properly, so let's add a `styles` prop and inject it in the document `head`.

EDIT:

besides that, I did a full component refactor:

- moving deprecated default proptypes to deconstruction default values
- converting class into function
- use hooks (useState, useMount, useEffect)

### DIFF
--- a/components/script/loader/README.md
+++ b/components/script/loader/README.md
@@ -20,7 +20,10 @@ $ npm install @schibstedspain/sui-script-loader --save
 ```js
 import ScriptLoader from '@schibstedspain/sui-script-loader'
 
-return (<ScriptLoader src={src} verifier={() => true} isAsync={false} render={() => 'Ready to render!'} />)
+const src = 'javascript file url'
+const styles = 'styles file url'
+
+return (<ScriptLoader src={src} verifier={() => true} isAsync={false} render={() => 'Ready to render!'} styles={styles}/>)
 ```
 
 **Find full description and more examples in the [demo page](https://sui-components.now.sh/workbench/script/loader).**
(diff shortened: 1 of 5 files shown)


midudev:
This seems to be misleading. `styles` seems to be a bunch of files or inline-styles. Yet is a single file. What about `stylesheet`?

JoanClaret:
@midudev totally agree.
fixed!

ruben-martin-lozano:
What about creating an `<StyleLoader />`?

desko27:
@ruben-martin-lozano 
> What about creating an `<StyleLoader />`?

I find it interesting. Although I think that the OP is thinking in a dual scenario where you load styles which are associated to a particular script, so you load them in the same `<ScriptLoader />`.

carlosvillu:
Maybe I lost something, but to do this, we have Helmet. I mean, We are using ScritpLoader because we need extra features like retries o callback function. But to inject a Link tag in the header, with Helmet should be enough

ruben-martin-lozano:
If so, can we use Helmet for adding this Link tag in the header instead of using ScriptLoader?

JoanClaret:
thanks for your feedback guys!
i get your point @carlosvillu to be honest, i didn't thought about to use the helmet. anyways, to me this way looks a bit more simple because the load of the script and styles are done in the same file.

besides that, i did a full component refactor:
- moving deprecated default proptypes to deconstruction default values
- converting class into function
- use hooks (useState, useMount, useEffect)

so, we can either simply close this PR and try to send the stylesheet via helmet, or test and merge this PR.

carlosvillu:
as you say is a fully refactor and that is worthy. you have my 👍

JoanClaret:
Great! Thank you!
I've tested this component in motor, but it's also used in FC:

https://github.schibsted.io/scmspain/frontend-fc--uilib-components/search?q=ScriptLoader&unscoped_q=ScriptLoader

Any volunteer to help me testing it in fc environment?

JoanClaret:
Awesome! thank you all, mates! 🙌 
Keep PR'ing ✊

PR #99, opened 2019-07-22
Feature/add use state reducer to react hooks

desko27:
## ✍️ Description

Hook that provides a similar state management approach to the old `this.setState` from a class.

Useful to cover quick functional migrations in components with complex states that would cause unnecessary renders if simply divided into lots of useState hooks.

If you apply this when migrating to a functional component, please take in mind that you should later rethink the strategy of its state.

## ☑️ Tasks

- [x] Add new `useLegacyState` hook
- [x] Add docs
- [x] Add demo


midudev:
I would change the name to `useLegacyState` in order to make it clear this behaviour is not recommended nor a good idea by default. :D

midudev:
You could remove the action for the reducer and use by default the one that do the reduce, to simplify the code.

Or even using `useState` directly should do the trick (not tested :D).

```javascript
import React, { useState } from 'react'

export default function (initialState = {}) {
  const [state, setState] = useState(initialState)
  const setLegacyState =  newState => { setState({ ...state, ...newState }) }

  return [state, setLegacyState]
}
```

PR #100, opened 2019-07-22
feat(map/basic): allow featureCollection maps to be rendered. 

ruben-martin-lozano:
Also, fix hovering styles on each polygon layer.


components/map/basic/src/leaflet/shapes/Polygons.js
```diff
@@ -19,62 +20,45 @@ export default class SearchMapPolygons {
...
-    const polygonGeoJSon = L.geoJson(polygon, params)
+      onEachFeature: (feature, layer) => {
+        layer.on({
+          mouseout: function() {
+            polygonGeoJSon.resetStyle(this)
+          },
+          mouseover: function() {
+            this.setStyle(hoverStyles)
+
+            if (!L.Browser.ie && !L.Browser.opera) {
```
AitorRodriguez990: Why this?

ruben-martin-lozano: tbh, don't know, it was already in the code and I've moved to another line ;)

PR #105, opened 2019-07-30
feat(map/basic): add markers logic

juanmendez90:
Add highlighted marker option, also change the priority of status to Contacted > Favorited > Visited


components/map/basic/src/leaflet/map.js
```diff
@@ -237,7 +237,11 @@ export default class LeafletMap {
       !arrayToCompare.some(comparedArrayPoint => {
```
midudev: A little bit nicer and readable:
```
const { Id, isSelected, propertyInfo } = comparedArrayPoint
const { isFavorite, highlighted } = propertyInfo
return Id === originalArrayPoint.Id && isSelected === originalArrayPoint.Id ...

PR #106, opened 2019-07-30
Feature/map pois update

juanmendez90:


components/map/basic/src/leaflet/marker-manager.js
```diff
@@ -173,16 +178,19 @@ class MarkerManager {
...
 
     const checkModifier = className => {
       return classModifiers[className](options)
     }
 
-    const modifier = Object.keys(classModifiers).find(checkModifier)
+    let modifier = Object.keys(classModifiers).find(checkModifier)
+    modifier = `${modifier} ${
+      this.isHighlighted(options) ? HIGHLIGHTED_CLASS : ''
+    }`
```
desko27: I'd use another const instead of mutating a var:
```js
const matchingModifier = Object.keys(classModifiers).find(checkModifier)
const modifier = `${matchingModifier} ${
  this.isHighlighted(options) ? HIGHLIGHTED_CLASS : ''
}`
```

PR #115, opened 2019-08-23
Feat/cmp get fixed

midudev:
New CMP version with fixes and new layout.

ezgif-5-3044ba33cb9a


components/cmp/modal/src/Consents/index.js
```diff
@@ -35,29 +38,52 @@ export const Consents = ({
...
+        }) => (
+          <ConsentItem
+            enabled={consents[id]}
+            id={id}
+            isVendor={isVendor}
+            key={id}
+            name={name}
+            renderDescription={() => (
+              <ConsentDescription
+                isVendor={isVendor}
```
juanmendez90: order of props

PR #121, opened 2019-09-13
feat(map/basic): do not show label to our current location

AitorRodriguez990:
## Description

Show label only on children and brothers. Until now, we were showing the label for the current location too.

## Preview

In this example, we removed "El Masnou" label and only show its children.

image

### DIFF
--- a/components/map/basic/src/leaflet/shapes/Polygons.js
+++ b/components/map/basic/src/leaflet/shapes/Polygons.js
@@ -84,9 +84,11 @@ export default class SearchMapPolygons {
           return
 
         try {
-          that.currentGeoCode !== layer.feature.properties.Code &&
+          const {Code, LocationName} = layer.feature.properties
+
+          that.currentGeoCode !== Code &&
             layer
-              .bindTooltip(layer.feature.properties.LocationName, {
+              .bindTooltip(LocationName, {
                 permanent: true,
                 direction: 'center',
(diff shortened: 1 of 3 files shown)


components/map/basic/src/leaflet/shapes/Polygons.js
```diff
@@ -67,17 +74,26 @@ export default class SearchMapPolygons {
...
-          layer
-            .bindTooltip(layer.feature.properties.LocationName, {
-              permanent: true,
-              direction: 'center',
-              className: fitBound ? 'is-fit-bound' : ''
-            })
-            .openTooltip()
+          const {Code, LocationName} = layer.feature.properties
+
+          that.currentGeoCode !== Code &&
```
ruben-martin-lozano: Pondría una prop (boolean), para decidir si mostrar o no el tooltip para el current code

midudev:
We keep feeding the component with logic intended to be used only for FC. All of this should be placed outside this component, ideally on the FC component side.

components/map/basic/src/leaflet/shapes/Polygons.js
AitorRodriguez990: Done!

AitorRodriguez990:
@midudev `properties` is not a Real Estate property, all about `layer`, `feature` or ` properties` are Leaflet properties.

PR #127, opened 2019-09-20
feat(cmp/ui): set text-align: center to sui-CmpUi-content

jordevo:
Not really a feature nor a fix, but a style change in order to ensure that the text on the content of the banner is properly displayed (`text-align: left`).

image


jordevo:
> 👍
> 
> To solve everything, you'll have to do the same PR to https://github.com/SUI-Components/schibsted-spain-components/tree/master/components/cmp/modal
> 
> Because it happens the same once the modal is opened @jordevo

you're right! thanks! just added the css rule for cmp modal as well...

tureey:
Care because if you modify 2 components in the same `PR` the `check:release` doesn't detect the changes of one of those. (or at least it happened to me) @jordevo

PR #129, opened 2019-09-30
feat(map/basic): added hover function

juanmendez90:
Fix hover functionality, now if you step in a poi the polygon maintains its background-color.
The problem I am trying to solve here is that when you hover a POI on the map the hover class of the polygon is lost

### DIFF
--- a/components/map/basic/src/leaflet/map.js
+++ b/components/map/basic/src/leaflet/map.js
@@ -70,7 +70,8 @@ export default class LeafletMap {
       onPolygonWithBounds,
       showLabels
     })
-    this.polygons.setPolygonsOnMap({map: this._map, polygons})
+    const geoJSON = this.polygons.setPolygonsOnMap({map: this._map, polygons})
+    this.layerManager.addLayersToGroup(geoJSON, 'geoJSON')
   }
 
   createMarkerManager(mapId) {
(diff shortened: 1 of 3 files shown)


components/map/basic/src/leaflet/map.js
```diff
@@ -162,7 +163,123 @@ export default class LeafletMap {
     }
   }
 
+  pnpoly(x, y, coords) {
```
midudev: Is this magic? What's this doing?

components/map/basic/src/leaflet/map.js
```diff
@@ -162,7 +163,123 @@ export default class LeafletMap {
...
+    this._map.on('mousemove', ({latlng}) => {
+      const geoJSON = this.layerManager.getLayers(
+        this.layerManager.layers.geoJSON
+      )
+      geoJSON.forEach((item, index) => {
+        const point = {coordinates: [latlng.lng, latlng.lat], type: 'Point'}
+        item.eachLayer(layer => {
+          const {_path: path = {}} = layer
+          const {classList} = path
+          classList && classList.remove('re-Mappage-polygon-hover')
```
midudev: This is strongly attached to Fotocasa needs.

midudev:
Can we have more context about the problem that this is trying to solve? Tbh, I'm hardly able to follow the code :D

components/map/basic/src/leaflet/map.js
juanmendez90: I don't know 🤣, this was copy-paste from here https://github.com/maxogden/geojson-js-utils/blob/aa31043f31cb65d573a29dfc9faef49f12702b7c/geojson-utils.js#L72

components/map/basic/src/leaflet/map.js
juanmendez90: the map it self is strongly attached to fotocasa need. Should I send these classes via props?

midudev:
Closing as other PR was used for fixing this.

PR #135, opened 2019-10-11
fix(react/hooks): fix useMount effect with clear effect

giodelabarrera:
Fix to allow cleaning effects on hook
Fix not to cheat the deps `[]` of React -> `// eslint-disable-line`

Test where I run tests to compare the two different `useMount`:
[https://codesandbox.io/s/test-usemount-k9y43](https://codesandbox.io/s/test-usemount-k9y43)
`<FirstContent />` uses the current `useMount` and when unmounted does not clean the effect while `<SecondContent />` using the corrected `useMount` does.

### DIFF
--- a/components/react/hooks/src/useMount/index.js
+++ b/components/react/hooks/src/useMount/index.js
@@ -7,6 +7,6 @@ export default function useMount(effect) {
       clean = await effect()
     }
     doEffect()
-    return () => clean()
+    return () => typeof clean === 'function' && clean()
   }, []) // eslint-disable-line
 }
--- a/components/react/hooks/src/useMount/index.js
+++ b/components/react/hooks/src/useMount/index.js
@@ -1,5 +1,12 @@
 import {useEffect} from 'react'
 
 export default function useMount(effect) {
-  useEffect(effect, [])
+  useEffect(() => {
+    let clean = () => {}
+    async function doEffect() {
+      clean = await effect()
+    }
+    doEffect()
+    return () => clean()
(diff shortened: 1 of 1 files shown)


rmoralp:
This change is going to break Motor useMount in some components because we are using async functions in `useMount` callback.
Have a look on https://github.com/SUI-Components/schibsted-spain-components/commit/70b94a1198c72cbf7f994efa9d61b31a02ec0a93#diff-f7d18fa5f3a7ae717f6b515094cd8f30

midudev:
The problem was the `effect()` execution wasn't returning the execution so the callback to clean effects is not returned to the effect.

```
function useSuiMount(effect) {
  useEffect(() => effect(), [])
}
```

As the reference is being executed directly I guess this is working as expected but this could be problematic for `async` functions and this should be tested accordingly.

rmoralp:
useSuiMount works fine with async functions, but not works when return a function for "unMount"
This approach solves the second part but breaks the usage of async functions

midudev:
Still there's a minor bug :D When the effect is not returning anything. We should do instead:

```javascript
function useSUIMount(effect) {
  useEffect(() => {
    let clean = () => {};
    async function doEffect() {
      clean = await effect();
    }
    doEffect();
    return () => typeof clean === 'function' && clean();
  }, []); // eslint-disable-line
}
```

giodelabarrera:
It is already adapted to receive async methods and clean effects. 
It only remains to wait to leave it clean sometime...

PR #142, opened 2019-11-18
feat(perf/dynamicRendering): reuse hook for intersection observer and simplify code

midudev:
Use `useNearScreen` own hook in order to simplify perf dynamic rendering implementation.
Also, it will have the `intersection-observer` in a single place, perfect to keep in sync the version and avoid installing more than once.

### DIFF
--- a/components/perf/dynamicRendering/src/index.js
+++ b/components/perf/dynamicRendering/src/index.js
@@ -7,7 +7,7 @@ const BOTS_USER_AGENTS = [
   'google-structured-data-testing-tool',
   'bingbot',
   'linkedinbot',
-  'mediapartners-google',
+  'mediapartners-google'
 ]
 
 function checkUserAgentIsBot(userAgent, botsUserAgents) {
@@ -24,7 +24,7 @@ export default function PerfDynamicRendering({
   height = 0,
   placeholder,
   rootMargin,
-  userAgent,
+  userAgent
 }) {
   const isBot = checkUserAgentIsBot(userAgent, botsUserAgents)
   const isOnBrowser = typeof window !== 'undefined'
@@ -97,5 +97,5 @@ PerfDynamicRendering.propTypes = {
   /**
    * An Array of strings that is used to set the lists of userAgents for which the element is always rendered
    */
-  botsUserAgents: PropTypes.arrayOf(PropTypes.string),
+  botsUserAgents: PropTypes.arrayOf(PropTypes.string)
 }
--- a/components/perf/dynamicRendering/src/index.js
+++ b/components/perf/dynamicRendering/src/index.js
@@ -1,13 +1,13 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import {LazyContent} from './lazyContent'
+import LazyContent from './lazyContent'
 
 const BOTS_USER_AGENTS = [
   'googlebot',
   'google-structured-data-testing-tool',
   'bingbot',
   'linkedinbot',
-  'mediapartners-google'
+  'mediapartners-google',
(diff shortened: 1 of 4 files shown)


components/perf/dynamicRendering/src/lazyContent.js
```diff
@@ -1,73 +1,22 @@
...
+  height,
+}) {
+  const [isNearScreen, fromRef] = useNearScreen({offset: rootMargin})
+
+  if (isNearScreen) {
+    return children
+  } else if (placeholder) {
+    return <div ref={fromRef}>{placeholder}</div>
+  } else {
+    return <div ref={fromRef} style={{height: `${height}px`}} />
```
carlosvillu: I think that we need this hack here: `marginBottom = 1px`

components/perf/dynamicRendering/src/index.js
```diff
@@ -1,13 +1,13 @@
...
-import {LazyContent} from './lazyContent'
+import LazyContent from './lazyContent'
 
 const BOTS_USER_AGENTS = [
   'googlebot',
   'google-structured-data-testing-tool',
   'bingbot',
   'linkedinbot',
-  'mediapartners-google'
+  'mediapartners-google',
```
jcger: to early?

components/perf/dynamicRendering/src/lazyContent.js
midudev: The hack is already here: https://github.com/SUI-Components/schibsted-spain-components/pull/142/files#diff-11ae61defc028aa316a9a02f88c3c488R54

components/perf/dynamicRendering/src/index.js
midudev: :D I was testing the previous version. Fixed!

components/perf/dynamicRendering/src/lazyContent.js
```diff
@@ -1,73 +1,22 @@
...
+
+export default function LazyContent({
+  children,
+  rootMargin = '100px 0px 0px',
+  placeholder,
+  height
+}) {
+  const [isNearScreen, fromRef] = useNearScreen({offset: rootMargin})
+
+  if (isNearScreen) {
```
tureey: A bit late, but instead of :
```
if (isNearScreen) {
    return children
  } else if (placeholder) {
    return <div ref={fromRef}>{placeholder}</div>
  } else {
    return <div ref={fromRef} style={{height: `${height}px`}} />
  }
```

What do you think about:
```
if (isNearScreen) {return children}

if (placeholder) {
    return <div ref={fromRef}>{placeholder}</div>
}

return <div ref={fromRef} style={{height: `${height}px`}} />
```

Since we are executing a `return`, the `else statement` only adds mental complexity when evaluating the cases. @midudev

components/perf/dynamicRendering/src/lazyContent.js
```diff
@@ -1,73 +1,22 @@
...
-    }
+import {useNearScreen} from '@schibstedspain/sui-react-hooks'
+
+export default function LazyContent({
+  children,
+  rootMargin = '100px 0px 0px',
+  placeholder,
+  height
+}) {
+  const [isNearScreen, fromRef] = useNearScreen({offset: rootMargin})
```
tureey: super clean 👏

components/perf/dynamicRendering/src/lazyContent.js
midudev: I avoided to touch the return statement from the render in order to focus only on the things I wanted to change. :)

PR #145, opened 2019-11-22
Feature/pi27257

PablitoGS:


components/thumbnail/basic/src/index.scss
```diff
@@ -39,4 +41,11 @@
       max-height: 100%;
     }
   }
+
+  picture {
```
jcger: we should never apply styles on tags directly. Moreover, there is no picture tag in this component, might you be adding styles to another component? In this case, you should apply this styles there, maybe you need to add react props in order to make it work

PR #146, opened 2019-12-11
feat(cmp/ui): do not break on getvendors error

jcger:
Users are complaining that the banner does not close after clicking on "seguir navegando"

### DIFF
--- a/components/cmp/ui/src/CmpUi/index.js
+++ b/components/cmp/ui/src/CmpUi/index.js
@@ -31,12 +31,11 @@ export default function CmpUiContainer({
 
   const _handleAccept = async evt => {
     evt && evt.preventDefault()
-
-    const {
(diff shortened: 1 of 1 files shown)


components/cmp/ui/src/CmpUi/index.js
```diff
@@ -31,12 +31,11 @@ export default function CmpUiContainer({
 
   const _handleAccept = async evt => {
     evt && evt.preventDefault()
-
-    const {
-      purposeConsents,
-      vendorConsents
-    } = await getPurposesAndVendors.execute()
-    await sendConsents.execute({purposeConsents, vendorConsents})
+    try {
```
midudev: You need to send the consents you're getting still. The line 39 is missing and is needed.

PR #153, opened 2020-01-09
feat(gigya/globalEvents): move class to function and remove componentWillMount usage

midudev:
Use a function instead a class in order to remove the componentWillMount life cycle and simplify the component.

### DIFF
--- a/components/gigya/globalEvents/src/index.js
+++ b/components/gigya/globalEvents/src/index.js
@@ -1,4 +1,4 @@
-import {Component} from 'react'
+import {useEffect} from 'react'
 import PropTypes from 'prop-types'
 import {
   initListener,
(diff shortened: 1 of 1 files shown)


components/gigya/globalEvents/src/index.js
```diff
@@ -8,34 +8,24 @@ import {
...
     onLogout && removeOnLogoutSubscriber(onLogout)
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) { // eslint-disable-line
-    this._unregisterEvents(this.props)
-    this._registerEvents(nextProps)
-  }
-
-  UNSAFE_componentWillMount() { // eslint-disable-line
+  useEffect(function() {
```
giodelabarrera: He visto que haz hecho esto en otros useEffects, el usar una declaración de función y una expresión con arrow function. Es por alguna ventaja o gusto personal? 🙂

components/gigya/globalEvents/src/index.js
```diff
@@ -8,34 +8,24 @@ import {
...
     initListener()
-    this._registerEvents(this.props)
-  }
-
-  componentWillUnmount() {
-    this._unregisterEvents(this.props)
-  }
+    registerEvents({onLogin, onLogout})
```
giodelabarrera: Maybe you can move the functions `registerEvents` and `unregisterEvents` inside the useEffect. 
Didn't you skip the linter or have you wanted to ignore the deps of useEffect in an intentional way?

components/gigya/globalEvents/src/index.js
midudev: Gusto personal. :D

components/gigya/globalEvents/src/index.js
midudev: I'm just keeping the functionality of before that was using the `componentWillReceiveProps` without any check in order to register again the events.

PR #158, opened 2020-01-20
[form/builder] Move FormBuilder from MA Repo

carlosvillu:
This change is [Reviewable](https://reviewable.io/reviews/sui-components/adevinta-spain-components/158)


components/form/builder/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@s-ui/form-builder",
```
midudev: To follow the same package name we're using, this should be `@s-ui/react-form-builder`

components/form/builder/src/Icons/IconClose.js
```diff
@@ -0,0 +1,35 @@
+import React, {memo, useEffect, useState} from 'react'
+
+var SVGComponent = function SVGComponent(_ref) {
```
midudev: This file among other icons are unnecesarily already compiled for the src.

components/form/builder/README.md
```diff
@@ -0,0 +1,23 @@
+# FormBuilder
+
+> Description
+
+<!-- ![](./assets/preview.png) -->
+
+## Installation
+
+```sh
+$ npm install @s-ui/form-builder --save
```
davidmartin84: component name is: @s-ui/react-form-builder

components/form/builder/src/Icons/IconClose.js
carlosvillu: It is because I don't want to use an Iconset.

midudev: You don't need an iconset, you only have to avoid putting compiled files in the `src` folder and create a componente with the `svg` you want to use:

It would be something like this:
```
export default memo(() => (<svg viewBox='0 0 64 64'><path d='M60 62a2 2 0 0 1-1.41-.59L32 34.83 5.41 61.41a2 2 0 0 1-2.83-2.83L29.17 32 2.59 5.41a2 2 0 0 1 2.82-2.82L32 29.17 58.59 2.59a2 2 0 0 1 2.83 2.83L34.83 32l26.58 26.59A2 2 0 0 1 60 62z'></path></svg>)
))

File: components/form/builder/src/Input/index.js [hunk 4]
```diff
@@ -0,0 +1,139 @@
...
+  )
+
+  let nextProps = {}
+  switch (input.type) {
+    case FIELDS.TEXT: {
+      nextProps = {type: DISPLAY[input.display] || 'text'} // this is the html input type (not the field type related to DSL form-builder)
+
+      const constraints = input.constraints || []
+
+      nextProps = constraints.reduce((acc, constraint) => {
```
midudev: This kind of reduce and way of creating the object is very hard to read.

You might could consider something like this.

```
nextProps = constraints.reduce((acc, constraint) => {
  const {maxlength: maxLength, minlength: minLength, notnull, pattern} = constraint?.property?

  return {
    ...acc,
    ...(maxLength && { maxLength }),
    ...(minLength && { minLength }),
    ...(pattern && { pattern }),
    ...(notnull === '' && {
      required: true
    }),
  }
})
```

midudev: Also, it should benefit the codebase, in order to avoid doing more than once the optional chaining transformation.

components/form/builder/src/Input/index.js
```diff
@@ -0,0 +1,139 @@
...
+    value: input.value || '',
+    onChange: onChangeCallback,
+    size,
+    leftAddon,
+    rightAddon,
+    ...nextProps
+  }
+
+  if (!nextProps.type) {
+    return <span>Unkown Input type {input.type}</span>
```
midudev: Typo: `Unkown` -> `Unknown`

components/form/builder/src/Input/index.js
carlosvillu: I am going to merge and evaluate these optimizations in another PR.

PR #160, opened 2020-01-21
feat(card/arrow): refactor card arrow component allow optional elements

ivanmlaborda:
- Refactor card/arrow component to allow optional elements like image and description
- Add border-radius scss variable with default value to 0

*Desktop*
imagen

---

*Mobile*
imagen

### DIFF
--- a/components/card/arrow/src/index.js
+++ b/components/card/arrow/src/index.js
@@ -14,12 +14,16 @@ const CardArrow = ({media, text, icon, linkFactory: Link, link}) => {
   return (
     <div className="sui-CardArrow">
       <Link className="sui-CardArrow-link" href={link}>
-        <div className="sui-CardArrow-img">
-          <AtomImage src={media.src} alt={media.alt} />
-        </div>
+        {media && (
+          <div className="sui-CardArrow-img">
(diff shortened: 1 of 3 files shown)


components/card/arrow/src/index.js
```diff
@@ -14,12 +14,16 @@ const CardArrow = ({media, text, icon, linkFactory: Link, link}) => {
   return (
     <div className="sui-CardArrow">
       <Link className="sui-CardArrow-link" href={link}>
-        <div className="sui-CardArrow-img">
-          <AtomImage src={media.src} alt={media.alt} />
-        </div>
+        {media && (
+          <div className="sui-CardArrow-img">
+            <AtomImage src={media.src} alt={media.alt} />
```
tureey: I know its not a PR related thing, but what about adding the title attribute?

ivanmlaborda: well, in other cards maybe makes more sense, here I doubt that a tiny image has a SEO value.

tureey: well, it has not a role='presentation' or alt=''. This means this image is considered important for the content. Because of this, a title will help @ivanmlaborda

PR #164, opened 2020-01-22
feat(collapsible/showmore): new component "Show more"

nucliweb:
# In this Pull Request

CollapsideShowmore is a component to show the rest of the results in a list after pressed the **Show more** button and the button will disappear, we don't need a button "Show less" and is not for a pagination use

## Preview

preview

## Demo

preview


components/collapsible/showmore/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+import Button from '@s-ui/react-atom-button'
+import cx from 'classnames'
+
+const CollapsibleShowmore = ({
+  children,
+  copy = 'Show more',
+  maxHeight = 100
+}) => {
+  const baseClass = 'sui-CollapsibleShowmore'
+  const [heightContainer, setHeightContainer] = useState(maxHeight)
```
tureey: What about changing the order of the words to `containerHeight` and `setContainerHeight`?

components/collapsible/showmore/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+  const [hideButton, setHideButton] = useState(false)
+
+  const buttonClassName = cx(`${baseClass}-actionButton`, {
+    'is-hidden': hideButton
+  })
+
+  const unfold = () => {
```
tureey: 👏

components/collapsible/showmore/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+    setHeightContainer('auto')
+    setHideButton(true)
+  }
+
+  return (
+    <div className={`${baseClass}`}>
+      <div
+        className={`${baseClass}-container`}
+        style={{height: heightContainer}}
```
tureey: just so this doesn't grow, what about creating a variable? IT would then improve the readability of the render method. (but it's a totally personal opinion)

demo/collapsible/showmore/playground
```diff
@@ -0,0 +1,113 @@
...
+      </tr>
+      <tr>
+        <td>validationText</td>
+        <td>Atom</td>
+      </tr>
+    </tbody>
+  </table>
+)
+const children = <Table />
+const copy = 'Ver más'
```
tureey: As standard, constants must be in a capitalized snake case. For example:
```
const COPY
const ANOTHER_CONSTANT
```

demo/collapsible/showmore/playground
```diff
@@ -0,0 +1,113 @@
...
+const maxHeight = 210
+
+return (<CollapsibleShowmore 
+          children={children}
```
tureey: First time I see this. Just curious, is it the same as `<Component><Table/></Component>` isn't it?

jordevo: Yes. It probably makes more sense when the `CHILDREN` const is made up of multiple components though.

nucliweb: I want pass a component as a children, and in this demo I created a simple Table component.

components/collapsible/showmore/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "npx rimraf ./lib && npx mkdirp ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@s-ui/component-dependencies": "1",
+    "@s-ui/react-atom-button": "^1.38.0"
```
midudev: Use the major version of the package.

components/collapsible/showmore/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+  const baseClass = 'sui-CollapsibleShowmore'
+  const [heightContainer, setHeightContainer] = useState(maxHeight)
+  const [hideButton, setHideButton] = useState(false)
+
+  const buttonClassName = cx(`${baseClass}-actionButton`, {
+    'is-hidden': hideButton
+  })
+
+  const unfold = () => {
+    setHeightContainer('auto')
```
midudev: The `setHeightContainer` state seems to be unnecessary as could be calculated from the `hideButton` state.

components/collapsible/showmore/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+    setHideButton(true)
+  }
+
+  return (
+    <div className={`${baseClass}`}>
```
midudev: From: `<div className={`${baseClass}`}>`
To: `<div className={baseClass}>`

components/collapsible/showmore/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+        className={`${baseClass}-container`}
+        style={{height: heightContainer}}
+      >
+        {children}
+      </div>
+      <Button
+        className={buttonClassName}
+        design="outline"
+        title={copy}
+        onClick={unfold}
```
midudev: Is not linter complaining about this? Normally, you should use a `handle...`

demo/collapsible/showmore/playground
midudev:
```
<CollapsibleShowmore copy={copy} maxHeight={maxHeight}>
  {children}
</CollapsibleShowmore>
```

nucliweb:
Finally we will use the **SUI Collapsible** component current version to publish the page, and then we will iterate to add button to SUI Collapsible component.

cc @turolopezsanabria

PR #169, opened 2020-01-28
fix(form/builder): 1 ms setTimeout to avoid IE 11 error - Access is denied - XMLHttpRequest

tomasmax:
1 ms setTimeout to avoid IE 11 error - Access is denied - XMLHttpRequest


tomasmax:
😭

tomasmax:
https://stackoverflow.com/questions/26891783/ie-11-error-access-is-denied-xmlhttprequest

PR #170, opened 2020-01-29
fix(topbar/user): make sure dom elements are ready to make calculations from their style properties

kikoruiz:


components/topbar/user/src/index.js
```diff
@@ -41,25 +40,6 @@ export default function TopbarUser({
   const [navWrapStyle, setNavWrapStyle] = useState(DEFAULT_NAV_WRAP_STYLE)
   let _windowWidth
```
midudev: I think this should be a reference in order to avoid losing the value between renders.

PR #171, opened 2020-01-29
Feat/form builder validations

arnau-rius:
- Add fields validation using validity native api for a single fields or all fields.
- Fix a super weird performance bug with inputs w/ @carlosvillu .
- replace window hack key by useRef hook.


components/form/builder/src/Standard/index.js
```diff
@@ -39,12 +39,11 @@ const CONSTRAINTS = {
   minlength: 'tooShort'
 }
 
-const checkConstrainstsFactory = json => ({for: fieldID, all}) => {
-  const field = pickFieldById(json.form.fields, fieldID)
+const checkConstraintsFromField = field => {
   let errorMessages = []
   const constraints = field.constraints || []
+  const elementNode = document.getElementById(field.id)
   const elementValidity = document.getElementById(field.id)?.validity
```
midudev: Now you could use the `elementNode` here instead getting it again.

components/form/builder/src/Standard/index.js
```diff
@@ -92,8 +94,34 @@ const checkConstrainstsFactory = json => ({for: fieldID, all}) => {
...
+    fieldsToValidate = fieldsNamesInOrderOfDefinition(json?.form?.fields)
+  } else if (fieldID) {
+    fieldsToValidate = [fieldID]
+  } else {
+    window.console.warn(
+      '[form/builder]: checkConstrainstsFactory: Specify if you want to validate a specific field or all the fields'
+    )
+  }
+
+  let fieldsWithErrors = {}
```
midudev: If you're going to use the reduce like this, better use a `forEach`.

arnau-rius: because forEach is fastest or more redeable?

midudev: More readable and, in this case, avoid unnecesary transpilation with rest operator.

components/form/builder/src/Standard/index.js
```diff
@@ -92,8 +94,31 @@ const checkConstrainstsFactory = json => ({for: fieldID, all}) => {
     if (checkboxShouldBeTrueConstraint && !checkboxValue)
       errorMessages = [checkboxShouldBeTrueConstraint.message, ...errorMessages]
   }
-
   return errorMessages
 }
 
+const checkConstrainstsFactory = json => ({for: fieldID, all}) => {
+  let fieldsToValidate = {}
```
midudev: `fieldsToValidate` is initialized as an object but later seems to be an array. Is this correct? Because the `.reduce` or `forEach` will be able to run only on arrays.

arnau-rius: thanks, first draft implementation was an object..

components/form/builder/src/index.js
```diff
@@ -21,11 +25,9 @@ const FormBuilder = ({
 }) => {
   const {fields = [], rules = {}, id: formID} = json.form
   const [stateFields, setStateFields] = useState(fields)
-  const HACK_KEY = `__PERFORMANCE_UGLY_HACK_STATE_FIELDS_${formID}__`
   const [stateShowSpinner, setStateShowSpinner] = useState(false)
-  if (typeof window !== 'undefined') {
-    window[HACK_KEY] = stateFields
-  }
+  const __PERFORMANCE_UGLY_HACK_STATE_FIELDS__ = useRef()
```
quinwacca: 💘

PR #172, opened 2020-01-29
Feat/memoize form builder

arnau-rius:
##### What does this PR do?
- memoize components, so avoiding full render of all form components if a field changes
- abstract one function in index.js of form builder (DRY).
#####  Before
before

##### After
after


arnau-rius:
> A good PR! I did not know we could add customization to the comparison in the `react.memo()`
> 
> Good job! 👏

In fact if we use the default function in the React.memo it doesn't work as expected in some cases (for example if you have objects with several child deep levels).
By default React.memo() does a shallow comparison of props and objects of props:
More precisely this one: https://github.com/facebook/react/blob/v16.8.6/packages/shared/shallowEqual.js📚

carlosvillu:
Memo FTW !!!

components/form/builder/src/Checkbox/index.js
```diff
@@ -72,4 +72,11 @@ Checkbox.propTypes = {
   errors: PropTypes.object
 }
 
-export default Checkbox
+export default React.memo(Checkbox, (nextProps, prevProps) => {
+  return (
+    JSON.stringify(nextProps.checkbox) === JSON.stringify(prevProps.checkbox) &&
```
midudev: Shouldn't we extract this to a function in order to reuse in all components?

```
export const createComponentMemo = (field) => (nextProps, prevProps) => {
  return (
    JSON.stringify(nextProps[field]) === JSON.stringify(prevProps[field]) &&
      nextProps.errors[nextProps[field].id] === prevProps.errors[prevProps[field].id] &&
      nextProps.onChange === prevProps.onChange
})

export default React.memo(Checkbox, createComponentMemo('checkbox'))
```

PR #184, opened 2020-02-12
feat(form/builder): add reset state and force update from outside

arnau-rius:
- autosuggest field: if force value from outside sync state.
- checkbox field, uncomment styles.

### DIFF
--- a/components/form/builder/src/Select/Autosuggest/index.js
+++ b/components/form/builder/src/Select/Autosuggest/index.js
@@ -7,18 +7,19 @@ import MoleculeAutosuggestField from '@s-ui/react-molecule-autosuggest-field'
 import MoleculeAutosuggestOption from '@s-ui/react-molecule-dropdown-option'
 import IconClose from '../../Icons/IconClose'
 
-const AutosuggestSelect = ({select, tabIndex, onChange, size, errors}) => {
-  const fromValueToText = datalist => value => {
-    const item = datalist.find(item => item.value === value)
-    return item?.text
-  }
+const fromValueToText = datalist => value => {
+  const item = datalist.find(item => item.value === value)
+  return item?.text
+}
 
-  const fromTextToValue = datalist => text => {
-    const item = datalist.find(
-      item => item.text.toLowerCase() === text.toLowerCase()
-    )
-    return item?.value
-  }
+const fromTextToValue = datalist => text => {
+  const item = datalist.find(
+    item => item.text.toLowerCase() === text.toLowerCase()
+  )
+  return item?.value
+}
+
+const AutosuggestSelect = ({select, tabIndex, onChange, size, errors}) => {
   const errorMessages = errors[select.id]
   const {datalist = []} = select
   const fromTextToValueWithDatalist = fromTextToValue(datalist)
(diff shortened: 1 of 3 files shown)


components/form/builder/src/index.scss
```diff
@@ -1,10 +1,10 @@
 @import '~@schibstedspain/sui-theme/lib/index';
 @import '~@s-ui/react-atom-spinner/lib/index';
 @import '~@s-ui/react-atom-button/lib/index';
-@import './Checkbox/index';
 @import './Input/index';
 @import './Switch/index';
 @import './Select/index';
+@import './Checkbox/index';
```
arnau-rius: Select styles are resetting some checkbox styles, so we need checkbox after select import.

components/form/builder/src/Select/Autosuggest/index.js
```diff
@@ -24,13 +24,22 @@ const AutosuggestSelect = ({select, tabIndex, onChange, size, errors}) => {
...
+  const [localStateText, setLocalStateText] = useState('')
+
+  // case: a value is forced from outside the component
+  // if text from prop is different from current local state text, update local text state
+  if (textFromValueProp && textFromValueProp !== localStateText)
+    setLocalStateText(textFromValueProp)
+
+  // case: a reset is forced from outside the component
+  // if value from prop is 'reset', update local text state to empty
+  if (select.value === 'reset' && localStateText !== '') setLocalStateText('')
```
carlosvillu: Where came from this string "reset" ?!

components/form/builder/src/Select/Autosuggest/index.js
```diff
@@ -24,13 +24,18 @@ const AutosuggestSelect = ({select, tabIndex, onChange, size, errors}) => {
...
   const fromTextToValueWithDatalist = fromTextToValue(datalist)
   const fromValueToTextWithDatalist = fromValueToText(datalist)
-  const text = fromValueToTextWithDatalist(select.value)
-  const [stateText, setStateText] = useState(text)
+  const textFromValueProp = fromValueToTextWithDatalist(select.value)
+  const [localStateText, setLocalStateText] = useState('')
+
+  // case: a value is forced from outside the component
+  // if text from prop is different from current local state text, update local text state
+  if (textFromValueProp && textFromValueProp !== localStateText)
```
midudev: This is going to cause a re-render of the component the first time if the `textFromValueProp` is different to `''`.

arnau-rius: it is exactly what I want. Because If I want to set a text value from outside the component (textFromValueProp) I will need a re-render in order to print this textFromValueProp in the component

arnau-rius: I know this sync is kind of weird, but form builder works with the numerical ids and the autosuggest works with the text that is in the input...

midudev: I was only noticing it :). Normally this is a smell that something is not being handled correctly in React... I dunno if maybe you could try to approach the problem by using a `useEffect` to be fired only when the dependency is changed and avoid executing it in the first render 🤔 .

PR #188, opened 2020-02-20
Feature/add open events to topbar composable

desko27:
## 🎯 Why?

Some team at Real Estate needs events to be aware of `DropdownMenu` component opening or closing. FYI this PR follows the very same approach that was implemented here → SUI-Components/sui-components#991

## ➕ Bonus

Since we already have a `react/hooks` component within this repo I also added the `usePrevious` hook there.

### DIFF
--- a/components/react/hooks/src/usePrevious/index.js
+++ b/components/react/hooks/src/usePrevious/index.js
@@ -9,7 +9,7 @@ export default function usePrevious(value) {
   return ref.current
 }
 
-usePrevious.displayName = 'HookUseOnScreen'
+usePrevious.displayName = 'HookUsePrevious'
 usePrevious.propTypes = {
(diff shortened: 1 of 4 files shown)


components/react/hooks/src/usePrevious/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default function usePrevious(value) {
+  const ref = useRef()
+  useEffect(() => {
+    ref.current = value
+  }, [value])
+  return ref.current
+}
+
+usePrevious.displayName = 'HookUseOnScreen'
```
midudev: 👀 wrong display name here

desko27: Eeeeh 😱  thank you!!

PR #190, opened 2020-02-24
feat(react/hooks): useMediaQuery hook

rmoralp:
**useMediaQuery** Hook. We can use it combined with our vertical breakpoints in order to use browser mediaQueries (same as css) but in JS 😃 

It observes the document to detect when its media queries change, instead of polling the values periodically.
image

Support 
image


jordevo:
👏

PR #191, opened 2020-02-24
feat(react/hooks): added useScroll hook

nucliweb:
### useScrool

Hook to get the scroll position and the direction of scroll, limited to the Y-axis.

useScroll


nucliweb:
# UseCase

scroll

components/react/hooks/src/useScroll/index.js
```diff
@@ -0,0 +1,43 @@
+import {useState, useEffect} from 'react'
+
+export default function useScroll({axis = 'Y'} = {}) {
+  const getBounding = () =>
```
davecarter: I'd rather flip the condition here and use:
- a more specific naming
- optional chaining

```javascript

const getBoundingClientRect = () => 
   window?.document
        ? document.body.getBoundingClientRect()
        : 0 
```
I hope this returns the same result you are expecting 😃 !

nucliweb: PR updated, thank you @davecarter  for your suggestion 😊

components/react/hooks/src/useScroll/index.js
```diff
@@ -0,0 +1,41 @@
+import {useState, useEffect} from 'react'
+
+export default function useScroll({axis = 'Y'} = {}) {
+  const getBoundingClientRect = () =>
+    window?.document ? document.body.getBoundingClientRect() : 0
+
+  const AXIS_MAP = {
```
midudev: You could move this outside the component.

components/react/hooks/src/useScroll/index.js
```diff
@@ -0,0 +1,41 @@
+import {useState, useEffect} from 'react'
+
+export default function useScroll({axis = 'Y'} = {}) {
+  const getBoundingClientRect = () =>
+    window?.document ? document.body.getBoundingClientRect() : 0
```
midudev: This is going to break SSR as window is not defined. Optional chaining could protect your from an `undefined` value but not a not defined variable.

nucliweb: Ok, so is better with the first version , right?

👉🏼 https://github.com/SUI-Components/adevinta-spain-components/pull/191/files/413a52956fc2ef37bfdca1496c909d7fade0f141#diff-cd45a1a1bf87d161bc518751a0aaba1b

components/react/hooks/src/useScroll/index.js
```diff
@@ -0,0 +1,41 @@
...
+    const currentPosition = -currentBounding[boundingKey]
+
+    setScroll(({position: prevPosition}) => ({
+      bodyOffset: currentBounding,
+      position: currentPosition,
+      scrollDirection: prevPosition > currentPosition ? 'DOWN' : 'UP'
+    }))
+  }
+
+  useEffect(() => {
```
midudev: We could use `useMount` to avoid having the empty dependency with the disable of the linter.

components/react/hooks/src/useScroll/index.js
```diff
@@ -0,0 +1,41 @@
...
+    position: getBoundingClientRect()[boundingKey],
+    scrollDirection: null
+  })
+
+  const onScroll = e => {
+    const currentBounding = getBoundingClientRect()
+    const currentPosition = -currentBounding[boundingKey]
+
+    setScroll(({position: prevPosition}) => ({
+      bodyOffset: currentBounding,
```
midudev: The bodyOffset is not used anywhere but it's initialized in the state and later updated here. Can we just remove it?

components/react/hooks/src/useScroll/index.js
midudev: Yes!. You could move outside the function a const like this:
```
const isClient = typeof window != 'undefined' && window.document
```

And later use here:
```
const getBoundingClientRect = () => isClient ? document.body. getBoundingClientRect() : 0
```

PR #195, opened 2020-02-28
fix(image/slider): single image center position

AgonisticKatai:
### BEFORE
Captura de pantalla 2020-02-28 a las 10 45 18

### AFTER
Captura de pantalla 2020-02-28 a las 10 36 02

### DIFF
--- a/components/image/slider/src/index.js
+++ b/components/image/slider/src/index.js
@@ -64,6 +64,11 @@ export default function ImageSlider({
     sliderOptions.initialSlide || 0
   )
   const slides = getSlides(images, linkFactory)
+  const hasSingleImage = Boolean(slides.length === 1)
+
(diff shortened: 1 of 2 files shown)


components/image/slider/src/index.js
```diff
@@ -64,6 +64,11 @@ export default function ImageSlider({
     sliderOptions.initialSlide || 0
   )
   const slides = getSlides(images, linkFactory)
+  const hasSingleImage = Boolean(slides.length === 1)
```
midudev: `slides.length === 1` is already a boolean.

AgonisticKatai: It was to give a little more semantics, but yes, it obviously returns a Boolean.
I'll change it, thank you!

PR #209, opened 2020-03-19
feat(abtesting/optimizelyXExperiment): add EmptyVariation component

midudev:
In order to avoid noisy warning messages regarding using <Fragment> for variations, we're creating a new `EmptyVariation` that renders nothing but avoid warnings like this:

image

### DIFF
--- a/components/abtesting/optimizelyXExperiment/README.md
+++ b/components/abtesting/optimizelyXExperiment/README.md
@@ -43,7 +43,7 @@ If no variation is provided or OptimizelyX API is unavailable, `defaultVariation
 
 ### Optional: Empty variation
 
-If you need to put an empty variation, **do not use `<Fragment>` component as you will get a lot of warnings message in the console**. Instead, use the `<EmptyVariation>` component and pass the props of the experiment.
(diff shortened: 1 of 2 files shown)


components/abtesting/optimizelyXExperiment/README.md
```diff
@@ -41,6 +41,21 @@ Component will obtain OptimizelyX's decision for given `experimentId` and displa
 
 If no variation is provided or OptimizelyX API is unavailable, `defaultVariation` will be displayed.
 
+### Optional: Empty variation
+
+If you need to put an empty variation, **do not use `<Fragment>` component as you will get a lot of warnings message in the console**. Instead, use the `<EmptyVariation>` component and pass the props of the experiment.
```
jcger: maybe instead of "to put" you could use "to use"

I think it should be "a lot of warning messages" and "component and pass by the experiment's props"

midudev: Thanks!

midudev: Thanks! Fixed :)

PR #220, opened 2020-04-23
feat(list/bullet): apply changes to have responsive block & string

nucliweb:
# In this Pull Request

### Mobile version

mobile

### Desktop version

desktop

## Changes

This component has a responsive behavior and we need to add a new behavior for the new design, so we applied the necessary changes to get the new layout in the mobile version and maintain the compatibility, **the changes needs to update the component on Baller Landing** because we are changing the **BOOLEAN prop** `responsive`  to **STRING prop** `responsive='responsive'`

### Mobile version

mobile-new

### Desktop version

desktop-new

### DIFF
--- a/components/list/bullet/src/index.js
+++ b/components/list/bullet/src/index.js
@@ -4,10 +4,13 @@ import Bullet from './bullet.js'
 import cx from 'classnames'
 
 const BASE_CLASS = 'sui-ListBullet'
-
-const ListBullet = ({listItems, responsive, smallFont}) => {
+const RESPONSIVE = {
+  responsive: 'responsive',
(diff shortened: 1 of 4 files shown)


components/list/bullet/src/index.js
```diff
@@ -4,18 +4,27 @@ import Bullet from './bullet.js'
...
   })
 
   return (
     <div className={listBulletClass}>
       {listItems &&
         listItems.map((item, index) => (
-          <Bullet {...item} key={index} baseClass={BASE_CLASS} />
+          <Bullet
+            {...item}
+            isString={isString}
```
jordevo, line 24: 🍆 if it's a boolean, this could just be `isString`, couldn't it?

PR #228, opened 2020-05-06
fix(form/builder): fix onChange behaviour

AgonisticKatai:
### GOAL
When using version 2 of **FormBuilder** on the platform to test the **JSON** (workbench), an unexpected behavior has been observed, since it does not launch the **onChange** method at the expected time.
In one of the last commits, a duplicate call to the **onChange** method was removed, but it was found that the other one worked correctly, so the change was made.

### DIFF
--- a/components/form/builder/src/index.js
+++ b/components/form/builder/src/index.js
@@ -67,7 +67,7 @@ const FormBuilder = ({
     clearTimeout(timerShowSpinner)
     setStateFields(nextStateFields)
     onChange({
-      ...fieldsToObject(nextFields),
+      ...fieldsToObject(nextStateFields),
       __FIELD_CHANGED__: id
     })
     setStateShowSpinner(false)
--- a/components/form/builder/src/index.js
+++ b/components/form/builder/src/index.js
@@ -40,10 +40,7 @@ const FormBuilder = ({
       {value: fieldValue}
     )
     setStateFields(nextFields)
-    onChange({
-      ...fieldsToObject(nextFields),
-      __FIELD_CHANGED__: fieldId
-    })
+
     return nextFields
   }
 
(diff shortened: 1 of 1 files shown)


components/form/builder/src/index.js
```diff
@@ -69,6 +66,10 @@ const FormBuilder = ({
 
     clearTimeout(timerShowSpinner)
     setStateFields(nextStateFields)
+    onChange({
+      ...fieldsToObject(nextFields),
```
arnau-rius, line 70: The line that was removed, was this one: https://github.com/SUI-Components/adevinta-spain-components/pull/225/files#diff-b44d4b0d39148c1e3df23e43a1c54244L73 Pay attention that the fields that are passed were: `nextStateFields` instead of `nextFields`.

arnau-rius: If passing `nextFields` instead of `nextStateFields` you are sending the 'old' params in the onChange callback instead of the new ones.

AgonisticKatai: uaau sorry, FIXED!

arnau-rius: In fact, the duplication of onChange call was made intentionally because the function `reducerWithRules` can take some time depending on the number of rules. So the component that uses the form builder can obtain some quick feedback first and then apply all the rules and send onChange feedback again. I approved last PR (removing duplicates) without remembering this intentional behavior but it seems that anybody has been affected.

Form builder is hard and with 0 tests, but we should consider (me first) start writting tests in this component so we have the security that we are not introducing breaking changes since it is used in production in Milanuncios or creating new majors when we are not sure that the changes can break exisiting implementations so we can test new PR's. Otherwise it will not be possible to introduce new changes in the future without having the security that we are not breaking things. What do you think?

PR #229, opened 2020-05-06
feat(form/builder): field side margins as default variable

AgonisticKatai:
Due to the need for **UX**, we need to define different margins than those proposed for the default **FormBuilder**.
With this change we give the possibility to `overwrite this variable in the theme `of each vertical

### DIFF
--- a/components/form/builder/src/index.scss
+++ b/components/form/builder/src/index.scss
@@ -16,6 +16,7 @@ $bg-form-builder: transparent !default;
 $p-form-builder: 0 !default;
 $bdrs-form-builder: 0 !default;
 $fz-form-builder-label: $fz-base !default;
+$m-form-builder-field: $m-m !default;
 
 // TODO: Remove after SUI comp allows hidden accessible labels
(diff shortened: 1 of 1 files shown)


components/form/builder/src/index.scss
```diff
@@ -52,8 +53,8 @@ $fz-form-builder-label: $fz-base !default;
     box-sizing: border-box;
     flex-grow: 1;
     margin-bottom: $m-l;
```
andresin87, line 55: Might set a variable for the margin bottom too. Its name can collision with the proposed

AgonisticKatai: @arnau-rius & @andresin87 
You're right. I'll upload a new proposal.
How about $m-form-builder-field-side?

arnau-rius: Maybe you can tokenize the whole margin property: `margin: $m-form-builder-field` so there is no need to change the variable and we have all 4 margins tokenized.

PR #232, opened 2020-05-12
feat(react/hooks): add useDebounce hook

adriadoz:
**Add useDebounce hook to Adevinta Spain Components**

This hook allows you to debounce any fast changing value. The debounced value will only reflect the latest value when the useDebounce hook has not been called for the specified time period. When used in conjunction with useEffect, as we do in the recipe below, you can easily ensure that expensive operations like API calls are not executed too frequently. 

From: https://usehooks.com/useDebounce/


midudev:
Faltaría la documentación, añadirlo en el README.

PR #233, opened 2020-05-14
feat(form/builder): customize sui components via JS

arnau-rius:
#### Description
Add **children** prop, that should be a function (called renderer inside form builder) that receives two arguments: id and innerProps. 

If this function returns: 
- A `component` it will be rendered [instead](https://github.com/SUI-Components/adevinta-spain-components/pull/233/files#diff-5831da6776fc2aa1c60a4984599677d3R80) of the sui-component.
- A plain JS `object`, this object will be [passed as props of the sui-component](https://github.com/SUI-Components/adevinta-spain-components/pull/233/files#diff-5831da6776fc2aa1c60a4984599677d3R90).

Example:

```js
<FormBuilder
                  json={JSONBuilder}
                  onChange={fields => updateFields(fields)}
                  onBlur={onBlur}
                  errors={errors}
                >
                  {({id, innerProps}) => {
                    const props = {
                      color: {
                        size: 'large'
                      }
                    }
                    return props[id]
                  }}
</FormBuilder>
```
In this example the field with ID color will render the component defined in the JSON with size="large".

#### BREAKING CHANGE:
remove hardcoded values like size="long" inside some components.

#### MIGRATING GUIDE:
- If you are using a type="picker" with display="" (default) or display="dropdown" or display="autosuggest" the size of select sui component will no longer be large by default and you need to pass it as explained in the example. [Code removed](https://github.com/SUI-Components/adevinta-spain-components/pull/233/files#diff-4107bf78a3f3e980a181a6e1c7d4e5a1L86)
- If you are using a type="numeric" with display="money" the rightAddon will be empty by default and you will need to pass it as explained in the example [Code removed](https://github.com/SUI-Components/adevinta-spain-components/pull/233/files#diff-d1c5de68dcbc3d4a8180c327352aebf7L17)
- If you are using a type="text" with display="multiline" the size will be empty by default and you will need to pass it as explained in the example [Code removed](https://github.com/SUI-Components/adevinta-spain-components/pull/233/files#diff-2ae6c39f35e097c4a5db5e7f4bb944cdL79)

### DIFF
--- a/components/form/builder/src/ProxyField/index.js
+++ b/components/form/builder/src/ProxyField/index.js
@@ -1,5 +1,3 @@
-import React from 'react'
-
 import PropTypes from 'prop-types'
 import {field} from '../prop-types'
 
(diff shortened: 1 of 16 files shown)


components/form/builder/src/ProxyField/index.js
```diff
@@ -18,32 +16,71 @@ const ProxyField = ({
...
+      })
       break
 
     default:
-      Field = (
-        <p>
-          Unknown Field type <span>{field.type}</span>
-        </p>
-      )
+      Field = renderer({
```
carlosvillu, line 73: Maybe we can put the `Unknown` message if the renderer function returns a plain object. Because we don't know what to do with that information.

PR #235, opened 2020-05-15
Feat/add label to radio & add help to textarea

arnau-rius:
Add label manually into `Radio` type since there no label into `MoleculeRadioButtonGroup` component.
Add help to `Textarea`


components/form/builder/src/TextArea/index.js
```diff
@@ -69,6 +69,7 @@ const TextArea = ({
     id: textArea.id,
     name: textArea.id,
     label: textArea.label || '',
+    help: textArea.help,
```
carlosvillu, line 72: "label" and "help" are defined in the DSL?

PR #236, opened 2020-05-19
Feature/abtesting backward compatible unique contexts by experiment name

desko27:
# Add unique contexts by experiment name
Solution is backward compatible. The adoption of the new setup is not mandatory but convenient since it removes an important limitation. AB Testing tools are heavily tested so nothing is broken.

- [x] Docs
- [x] Extra tests

About the method to keep multiple contexts under the hood, @midudev deserves kudos for pointing it out! 🎉 Thanks man!

## 🔴 Limitation

We cannot nest multiple experiment contexts because they're actually a single context instance.

#### Parent component
```jsx
const experimentDataX = useExperiment({
  experimentId: 40000,
  variations: [{id: 700000, isDefault: true}, {id: 700001}]
})

const experimentDataY = useExperiment({
  experimentId: 50000,
  variations: [{id: 900000, isDefault: true}, {id: 900001}]
})

return (
  <OptimizelyXExperiment feed={experimentDataX}>
    <OptimizelyXExperiment feed={experimentDataY}>
      {/* Only Y is available in descendants... */}
    </OptimizelyXExperiment>
  </OptimizelyXExperiment>
)
```

#### Descendant
```js
const experimentDataY = useExperiment() // only Y is available here
```

## ✅ Solution

We want to have all parent experiments at disposal to be consumed from any descendant. To do that, we need to be able to ask for a specific experiment from a descendant, so we cannot continue to not passing any arguments to `useExperiment` at all.

Proposed solution looks like this:

#### Parent component
```jsx
const experimentDataX = useExperiment({
  name: 'experiment-x',
  experimentId: 40000,
  variations: [{id: 700000, isDefault: true}, {id: 700001}]
})

const experimentDataY = useExperiment({
  name: 'experiment-y',
  experimentId: 50000,
  variations: [{id: 900000, isDefault: true}, {id: 900001}]
})

return (
  <OptimizelyXExperiment feed={experimentDataX}>
    <OptimizelyXExperiment feed={experimentDataY}>
      {/* Both X and Y available in descendants... */}
    </OptimizelyXExperiment>
  </OptimizelyXExperiment>
)
```

#### Descendant
```js
const experimentDataX = useExperiment('experiment-x')
const experimentDataY = useExperiment('experiment-y')
```

### DIFF
--- a/components/abtesting/optimizelyXExperiment/src/context.js
+++ b/components/abtesting/optimizelyXExperiment/src/context.js
@@ -1,0 +1,13 @@
+import {createContext} from 'react'
+
+const experimentContexts = {
+  root: createContext()
+}
+
+export function getExperimentContext(name = 'root') {
+  if (!experimentContexts[name]) {
+    // create new context for `name` and store it to `experimentContexts`
+    experimentContexts[name] = createContext()
+  }
+  return experimentContexts[name]
+}
--- a/components/abtesting/hooks/README.md
+++ b/components/abtesting/hooks/README.md
@@ -19,10 +19,10 @@ $ npm install @s-ui/abtesting-hooks --save
 #### ⎡🧪👂⎦ Experiment Context Consumer
 
 ```js
-const {isActive, isDefault, isVariation, ...} = useExperiment()
+const {isActive, isDefault, isVariation, ...} = useExperiment('myexperiment')
 ```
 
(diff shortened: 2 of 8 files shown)


components/abtesting/optimizelyXExperiment/src/context.js
```diff
@@ -0,0 +1,13 @@
+import {createContext} from 'react'
```
midudev, line 1: 💘

PR #238, opened 2020-05-21
feat(form/builder): add onchangeCallbacck inlinebutton formbuilder

davidmartin84:


components/form/builder/src/InlineButton/index.js
```diff
@@ -21,10 +21,19 @@ const InlineButton = ({
     return onChange(inlineButton.id, value)
   }
 
+  const onChangeCallback = useCallback(
+    value => {
+      return onChange(inlineButton.id, value)
+    },
+    [onChange, inlineButton]
+  )
+
```
arnau-rius, line 30: maybe we don't need this callback, just passing onChange: onChange in the `inlineButtonProps` object is enough.. what do you think?

PR #240, opened 2020-05-23
feat(abtesting/hooks): add forceExperiment browser query flag

desko27:
# Add forceExperiment browser query flag

Now that we can give names to experiments, it's easy to provide extra methods of fake activation out of the code scope.

This way, UX or other roles can check variations whenever they like without the need of setting up and publishing the AB Test on Optimizely Panel, which is slower, cumbersome and wastes traffic from the plan.

### How does it work?

Simply add the `forceExperiment` query param to the URL on the address bar, providing experiment name and variation you want to activate as a value:
```
?forceExperiment=<experiment-name>|<variation-name>
```

For example, an experiment whose name is `myexperiment` with a `biggerButton` variation would look like this:
```
?forceExperiment=myexperiment|biggerButton
```

Note that this gets stored into the session storage, so the variation will show up through the entire navigation process until the tab gets closed. This is intended to avoid unnecessarily repeating the process of adding the mentioned query param.

### DIFF
--- a/components/abtesting/hooks/src/useExperimentCore/index.js
+++ b/components/abtesting/hooks/src/useExperimentCore/index.js
@@ -116,6 +116,10 @@ export default params => {
       const getMessageForParam = param =>
         `[ExperimentCore] Watch out!! Optimizely response will be ignored because the ${param} param.`
 
+      /*
+       * [Fake activation]
+       * Activate variation via `forceActivation` API flag
+       */
       if (forceActivation) {
         setTimeout(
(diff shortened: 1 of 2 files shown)


desko27:
Thank you @midudev! I'll submit another PR very soon to make some refactoring to the docs, because same concepts are actually split through two different packages, which I think is an issue that's growing bigger.

desko27:
Added some docs, they're going to be reused anyway!

PR #241, opened 2020-05-26
feat(abtesting/hooks): sequence of activations force experiment query…

desko27:
An improvement over #240 which enables the capability of accumulating multiple experiment activations by saving `forceExperiment` query param action to a unique session storage key.

### DIFF
--- a/components/abtesting/hooks/src/useExperimentCore/index.js
+++ b/components/abtesting/hooks/src/useExperimentCore/index.js
@@ -164,8 +164,9 @@ export default params => {
           // so it can override session storage entry.
           const [nameFromQP] = splitNames(forceExperimentFromQueryParam)
           const [nameFromSS] = splitNames(forceExperimentFromSessionStorage)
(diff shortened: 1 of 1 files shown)


components/abtesting/hooks/src/useExperimentCore/index.js
```diff
@@ -146,14 +146,36 @@ export default params => {
...
+        if (
+          forceExperimentFromQueryParam &&
+          forceExperimentFromSessionStorage
+        ) {
+          // When both query param and session storage entry are present,
+          // query param has preference BUT only if it's the same name,
+          // so it can override session storage entry.
+          const [nameFromQP] = splitNames(forceExperimentFromQueryParam)
+          const [nameFromSS] = splitNames(forceExperimentFromSessionStorage)
+          if (nameFromQP === nameFromSS) return forceExperimentFromQueryParam
```
midudev, line 167: I think this would be more readable with a ternary.

PR #245, opened 2020-06-15
feat(list/bullet): added support Heading prop to improve the SEO

nucliweb:
## In this Pull Request

As an SEO requirement, we need to define the heading in the bullet title.

### DIFF
--- a/components/list/bullet/src/index.js
+++ b/components/list/bullet/src/index.js
@@ -1,20 +1,11 @@
 import PropTypes from 'prop-types'
 import React from 'react'
-import Bullet from './bullet.js'
+import Bullet, {headingTags} from './bullet.js'
 import cx from 'classnames'
 
 export const responsiveTypes = {
   responsive: 'responsive',
   responsiveBlock: 'responsiveBlock'
-}
-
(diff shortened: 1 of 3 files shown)


AgonisticKatai:
With the previous behavior, the headings of the image title and the inner title had different weight (h2 and h3). Now both have been defined with the same heading (h2).
This behaviour is the one we want?

components/list/bullet/src/index.js
```diff
@@ -8,9 +8,24 @@ export const responsiveTypes = {
...
 }
 
+const HEADING_TAGS = {
+  h1: 'h1',
+  h2: 'h2',
+  h3: 'h3',
+  h4: 'h4',
+  h5: 'h5',
+  h6: 'h6'
+}
```
jordevo, line 18: 🍆 couldn't you have them imported from `bullet.js`?

nucliweb:
> With the previous behavior, the headings of the image title and the inner title had different weight (h2 and h3). Now both have been defined with the same heading (h2).
> This behaviour is the one we want?

Yes, really the previous behavior was an error, the illustration and  el title will never be displayed at the same time

PR #247, opened 2020-06-17
Feature/tcf boros import

maryammdot:
## Description

This PR changes the boros mock used in the tcf services component for the real boros object imported. 
It also updates the firstLayer and the secondLayer components with the corresponding changes to work with the data structure that comes from boros mock. 
Finally it takes in account that the objects that come from the loadUserConsent method from boros can be empty, and it's responsibility of the first and second layer components to fill them with all the keys that have it's homolog object that came from the getVendorList method and with value true or false depending on the user action. 

## Solves ticket/s

https://jira.scmspain.com/browse/PSP-3199

## Expected behavior

The behavior has to be the same as before. 
There will be more data now, because Boros library also returns the special features node from the consent object.

## Further considerations

This PR has to be merged before the [PR](https://github.com/SUI-Components/adevinta-spain-components/pull/248) for the update of the tcf ui component. It had to be done in different PRs to avoid any issue with the publication of packages (because the ui component has as dependencies the firsLayer, secondLayer and services components)


components/tcf/firstLayer/src/index.js
```diff
@@ -32,15 +32,36 @@ export default function TcfFirstLayer({
   const handleSettingsClick = () => {
     openSecondLayer()
   }
+  const formatConsentObject = ({purpose = {}}) => {
+    if (!purpose.consents) {
+      purpose.consents = {}
```
alextremp, line 37: NBC, only if it seems better to you.. 
```
purpose.consents = purpose.consents || {}
purpose.legitimateInterests = purpose.legitimateInterests || {}
Object.keys(vendorListState.purposes).forEach(key => {
  purpose.consents[key] = !!purpose.consents[key]
  purpose.legitimateInterests[key] = !!purpose.legitimateInterests[key]
})
```

components/tcf/secondLayer/src/index.js
```diff
@@ -133,26 +196,46 @@ export default function TcfSecondLayer({
...
+              state={state.purposes}
+              onConsentsChange={handlePurposesConsentsChange}
+              onLegitimateInterestsChange={handlePurposesLIChange}
+              hasLegitimateInterest
+            />
+          )}
+        {state &&
+          state.vendors &&
+          vendorListState &&
+          vendorListState.vendors && (
```
paulusrex, line 216: NBC. You can use if you want
state?.vendors && vendorListState?.vendors && ...

components/tcf/secondLayer/src/index.js
```diff
@@ -133,26 +196,46 @@ export default function TcfSecondLayer({
...
+              state={state.vendors}
+              onConsentsChange={handleVendorsConsentsChange}
+              onLegitimateInterestsChange={handleVendorsLIChange}
+              hasLegitimateInterest
+            />
+          )}
+        {state &&
+          state.specialFeatures &&
+          vendorListState &&
+          vendorListState.specialFeatures && (
```
paulusrex, line 230: Same as above

maryammdot: done, thanks

components/tcf/secondLayer/src/index.js
maryammdot: done, thanks

components/tcf/firstLayer/src/index.js
maryammdot: done, thanks!

PR #249, opened 2020-06-25
Fature/tcf first layer 2o iteration

maryammdot:
## Description
Second iteration of the TCF First Layer following final mockup.
Add real behaviour of the buttons and links.
Add styling for desktop, tablet & mobile.

## Solves ticket/s
https://jira.scmspain.com/browse/PSP-3292

## Expected behavior
Desktop:
image
Tablet:
image
Mobile:
image

## Review steps
Use the demo and change with the given buttons the desktop/mobile/tablet display. Review all buttons actions and also the styling.

## Further considerations
This is not the final implementation. It needs styling and scss variables refactor, to use sui-theme variables and ensure correct styling in all sites.
The counter approximation to get the initial pageYOffset can be far from the real initial Y scroll point (in the case that it's in the top, so it should be 0) so the user will have to do more than 250px of scroll to accept the consent. The gap with the real Y initial point will only be big if the user scroll is very fast, and in that case, it will be more than 250px so the consent will be accepted in the first scroll. It's because of this that in my opinion there is no problem with this gap, but it can be discussed or refactored.


components/tcf/firstLayer/src/index.js
```diff
@@ -1,114 +1,204 @@
...
+      initialYOffset - window.pageYOffset <= -250
+    ) {
+      handleSaveExitClick()
     }
-    loadConsent()
-    loadVendorList()
-  }, [getVendorList, loadUserConsent])
+  }
+  useEffect(() => {
+    document.addEventListener('scroll', checkScroll)
```
alextremp, line 72: maybe using a passive event listener would be better

See https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md
And https://caniuse.com/#feat=passive-event-listener

components/tcf/firstLayer/src/index.js
```diff
@@ -1,114 +1,204 @@
...
+    },
+    [handleCookiePolicyLayerClick]
+  )
+
+  let initialYOffset
+  let count = 0
+  const checkScroll = () => {
+    count++
+    // use counter to avoid initial window.pageYOffset pointing erroneously to 0
+    if (count === 3) {
```
alextremp, line 61: why 3?

components/tcf/firstLayer/src/index.js
```diff
@@ -1,114 +1,204 @@
...
+      initialYOffset = window.pageYOffset
     }
-    const loadVendorList = async () => {
-      const {purposes} = await getVendorList()
-      setVendorListState({purposes})
+    if (
+      (count >= 3 && initialYOffset - window.pageYOffset >= 250) ||
```
alextremp, line 65: if 250 is the IAB requirement for scrolling, maybe it should be a declared `const` instead of a magic number (it's being used at least in 2 lines, that should change if tomorrow the IAB says 300 instead of 250)

components/tcf/firstLayer/src/index.js
```diff
@@ -1,114 +1,204 @@
...
+              <div
+                className={
+                  isTablet
+                    ? `${CLASS}-buttons ${CLASS}-buttons--footer ${CLASS}-buttons--tablet`
+                    : `${CLASS}-buttons ${CLASS}-buttons--footer`
+                }
+              >
+                <Buttons />
+              </div>
+            </div>
```
alextremp, line 186: IMO, this is a duplication that may be extracted to a component which accepts the `className` variations

components/tcf/firstLayer/src/index.js
maryammdot: Done, I don't know if is an IAB requirement or Adevinta requirement, but I've added it as a constant. Thanks

components/tcf/firstLayer/src/index.js
maryammdot: There are only two divs, I don't think that we need an extracted component to create a div with some classes. Do you think some function as classNameGeneration would be better? @alextremp

components/tcf/firstLayer/src/index.js
maryammdot: When I was developing it, in mobile, the first two times that checkScroll was executed the value was 0, and in the third time, the value was the real Y position. I've raised the demo again to check it again and now from thee first execution of that function the value is the real Y position. I've removed the counter code. I don't understand why was acting that way and now is not doing it again. We should be aware of that in future refactors.

components/tcf/firstLayer/src/index.js
maryammdot: If I've understood well the documentation it seems that this passive option has to be used for events that can delay the scrolling response, not for the scroll event itself. 

`By marking a touch or wheel listener as passive, the developer is promising the handler won't call preventDefault to disable scrolling.`

Don't you understand it this way?

maryammdot:
Do you think we should use a debounce for the scroll event listener? something like: 

`const debounce = (fn, time) => {
    let timeoutId
    return function() {
      clearTimeout(timeoutId)
      timeoutId = setTimeout(fn, time)
    }
  }

document.addEventListener('scroll', debounce(checkScroll, 100))`

components/tcf/firstLayer/src/index.js
alextremp: Well, all non-passive listeners delay the whatever thing they're listening to, as they cannot know if the listener will cancel the event propagation by calling the `preventDefault`, so they must always wait for listener execution to complete.
With passive event listeners, the "listener code" is ran (in some moment), but without blocking.
This is so small code, so it's OK for me to go as it is 👍  (just seemed nice to me to try this xD)

components/tcf/firstLayer/src/index.js
alextremp: No no, I refer to this: both inner HTML structure are the same

```
          <div className={`${CLASS}-body`}>
            <span
              className={`${CLASS}-body-info`}
              ref={textRef}
              dangerouslySetInnerHTML={{__html: i18n.BODY}}
            />
            <div className={`${CLASS}-buttons`}>
              <Buttons />
            </div>
          </div>
```

This has className variations depending on the device (desktop, mobile, tablet) but essentially, it's a structure that can be extracted to a component.

maryammdot: done, thanks

components/tcf/firstLayer/src/index.js
maryammdot: Done! thanks

PR #253, opened 2020-07-03
fix(form/builder): fix typo on function and export

jordevo:
Fixes a function name containing a typo. Keeps backwards compatibility.

### DIFF
--- a/components/form/builder/src/index.js
+++ b/components/form/builder/src/index.js
@@ -162,9 +162,5 @@ export {fieldSizes as formBuilderFieldSizes}
 export {pickFieldById as formBuilderPickFieldById}
 export {changeFieldById as formBuilderChangeFieldById}
 export {fieldsNamesInOrderOfDefinition as formBuilderFieldsNamesInOrderOfDefinition}
-export {
-  checkConstraintsFactory,
-  // WARNING do not use the following export, it is kept for backwards compatibility
-  checkConstrainstsFactory
-} from './Standard'
+export {checkConstraintsFactory, checkConstrainstsFactory} from './Standard'
 export default FormBuilder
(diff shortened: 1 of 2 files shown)


components/form/builder/src/index.js
```diff
@@ -162,5 +162,9 @@ export {fieldSizes as formBuilderFieldSizes}
 export {pickFieldById as formBuilderPickFieldById}
 export {changeFieldById as formBuilderChangeFieldById}
 export {fieldsNamesInOrderOfDefinition as formBuilderFieldsNamesInOrderOfDefinition}
-export {checkConstrainstsFactory} from './Standard'
+export {
+  checkConstraintsFactory,
+  // WARNING do not use the following export, it is kept for backwards compatibility
+  checkConstraintsFactory as checkConstrainstsFactory
```
carlosvillu, line 168: I think it's more interesting if you do a function that only calls the right one inside but first puts a warning on the console that it's deprecated. You can do that so that the console is only seen if the environment is not Production

jordevo: how about this https://github.com/SUI-Components/adevinta-spain-components/pull/253/commits/f2555aab5f6849435c7aaa5e63476c547922e917 ?

components/form/builder/src/index.js
```diff
@@ -162,5 +162,9 @@ export {fieldSizes as formBuilderFieldSizes}
 export {pickFieldById as formBuilderPickFieldById}
 export {changeFieldById as formBuilderChangeFieldById}
 export {fieldsNamesInOrderOfDefinition as formBuilderFieldsNamesInOrderOfDefinition}
-export {checkConstrainstsFactory} from './Standard'
+export {
+  checkConstraintsFactory,
+  // WARNING do not use the following export, it is kept for backwards compatibility
```
rmoralp, line 167: Given you have added a console.warn this comment is not required.

PR #256, opened 2020-07-06
feat(list/link): use new @s-ui/react-link-basic

midudev:
Surprisingly, list link used the old version of link/basic. With this change we're using the latest compatible version.

So, I'm moving this in order to avoid installing two different packages for the same thing.

image

### DIFF
--- a/components/list/link/src/index.js
+++ b/components/list/link/src/index.js
@@ -2,26 +2,43 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
-import LinkBasic from '@schibstedspain/sui-link-basic'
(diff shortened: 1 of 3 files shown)


components/list/link/src/index.js
```diff
@@ -2,26 +2,43 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
-import LinkBasic from '@schibstedspain/sui-link-basic'
+import LinkBasic from '@s-ui/react-link-basic'
+
+function ListLinkItem({displayInline, item, useReactRouterLinks}) {
```
midudev, line 7: Extracted the method to a component in order to perform better instead a render method.

PR #263, opened 2020-07-09
fix(tcf/secondLayer): save purpose and specialFeatures data

alextremp:
## Description

image

With current version, no `purpose` data is being saved, neither `specialFeatures` nor `legitimateInterests`.
This fix adds the data initialization and manipulation from the second layer.

## Solves ticket/s

* https://jira.scmspain.com/browse/PSP-3347

## Expected behavior

Data is populated and saved to the consent string

image

## Review steps

Can be tested in the demo, check the saved value with `window.__tcfapi('getTCData', 2, console.log)`

## Memetized description

mandatory

### DIFF
--- a/components/tcf/secondLayer/src/index.js
+++ b/components/tcf/secondLayer/src/index.js
@@ -57,14 +57,12 @@ export default function TcfSecondLayer({
         userConsent.vendor.legitimateInterests = {
           ...userConsent.vendor.consents
         }
-        userConsent.purpose.consents = format({
-          reference: purposes,
-          object: {},
-          value: true
+        userConsent.purpose.consents = {}
+        userConsent.purpose.legitimateInterests = {}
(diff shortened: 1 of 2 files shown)


components/tcf/secondLayer/src/index.js
```diff
@@ -54,6 +54,21 @@ export default function TcfSecondLayer({
           object: {},
           value: true
         })
+        userConsent.vendor.legitimateInterests = {
+          ...userConsent.vendor.consents
```
maryammdot, line 58: Shouldn't be also included the ADEVINTA_COLLECTED_CONSENTS purposes validation? Now it includes all available purposes, but maybe some day new purposes are added and rules are not changed.

alextremp: maybe, it needs a big refactor IMO. If it's fine to you, it's better to merge as it is now, and refactor all the state manipulation in another PR

PR #264, opened 2020-07-09
feat(tcf/firstLayer): small changes in HTML and CSS

Osgar:
## Description
- Modifying classes to avoid adding complexity with devices
- Reduce class nomenclature
- Correct HTML semantics
- Modifying styles by adding media query

## Solves ticket/s
https://jira.scmspain.com/browse/PSP-3348

## Expected behavior
Modificación del footer en desktop:
Before:
image
After:
image
Protección visual contra posibles resizes despues de identificar el dispositivo
Tablet
Before:
image
image
After:
image

Mobile
Before:
image
After:
image
image

## Review steps
Open the component demo
Small changes in HTML and CSS


components/tcf/firstLayer/src/index.js
```diff
@@ -105,23 +102,14 @@ export default function TcfFirstLayer({
   }
 
   const Content = ({modalType}) => {
```
maryammdot, line 104: You should remove the param modalType because is not used any more

Osgar: @maryammdot Done 👍

PR #279, opened 2020-07-21
feat(tcf/secondLayer): second layer refactor for MUST tasks

Osgar:
## Description
Solve all the design tasks to come out with the test in Fotocasa

## Solves ticket/s
https://jira.scmspain.com/browse/PSP-3374

## Review steps
**Before**
image
image
image

**After**
image
image
image

## Further considerations
This PR is related to another one necessary to start the Fotocasa test:
https://github.mpi-internal.com/scmspain/frontend-fc--uilib-components/pull/1544

### DIFF
--- a/components/tcf/secondLayer/src/index.js
+++ b/components/tcf/secondLayer/src/index.js
@@ -183,10 +183,6 @@ export default function TcfSecondLayer({
     })
   }
 
-  const Logo = React.memo(() => (
-    <img className={`${CLASS}-logo`} src={logo} alt="logo" />
-  ))
-
   const vendorExpandedContent = props => (
     <TcfSecondLayerVendorExpandedContent
       {...props}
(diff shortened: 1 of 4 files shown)


components/tcf/secondLayer/src/index.js
```diff
@@ -209,13 +205,12 @@ export default function TcfSecondLayer({
         isOpen={modalOpen}
         closeOnOutsideClick
         closeOnEscKeyDown
-        header={isMobile ? <Logo /> : false}
+        header={<img className={`${CLASS}-logo`} src={logo} alt="logo" />}
```
maryammdot, line 208: Even you have to render it always, I would let the React.memo, to ensure that the logo that you pass to the modal is always the same. So you can let the logo as a constant with react memo and pass the <Logo /> component in the prop header

Osgar: @maryammdot The use of React.memo is not necessary now since the image is always fixed

PR #283, opened 2020-07-21
Task/refactor tcf ui layers

alextremp:
## Description

tcf/firstLayer and tcf/secondLayer are not usable components (they have no sense out of the context of the tcf/ui component), so it has more sense to discontinue them as public components and move them as sub components into the tcf/ui context (we're doing the same in the second layer with the decision groups for example)

This also:
- eases the local development as we'll don't have problems witht the Consent context provider.
- lazy loading the layers still works

## Solves ticket/s

* https://jira.scmspain.com/browse/PSP-3397

## Expected behavior

No changes, just copy&paste of the layers

## Review steps

Used the local demo
`npm run dev tcf/ui`

## Further considerations

* It contains the last changes that will go with the components (in #282 - last commit https://github.com/SUI-Components/adevinta-spain-components/pull/282/commits/598173ac60bd64abc4936d91519129895cb33404)
## Memetized description

mandatory


alextremp:
closing after merging from master as it would remove a new added component 🤷‍♂️

PR #294, opened 2020-07-28
feat(topbar/user): added the possibility to render custom content. Ad…

devictoribero:
This component is only used for `Coches` and `Motos` right now.

- Added the possibility to render custom content on the right instead of the `navCTA`.
- Added tests

Screen Shot 2020-07-28 at 11 25 57

### DIFF
--- a/components/topbar/user/src/index.js
+++ b/components/topbar/user/src/index.js
@@ -26,6 +26,7 @@ export default function TopbarUser({
   navMain,
   navUser,
   navCTA,
+  customContent,
   linkFactory = ({href, className, children, target, title}) => (
     <a href={href} className={className} target={target} title={title}>
       {children}
@@ -179,11 +180,8 @@ export default function TopbarUser({
   const toggleMenuClassName = cx('sui-TopbarUser-toggle', {
     'has-notifications': hasNotifications
   })
-  const {
-    url: navCtaUrl,
-    text: navCtaText,
-    onClick: onCTAClick = () => {}
-  } = navCTA
+
+  const handleCTAclick = navCTA && navCTA.onClick
 
   return (
     <div ref={_topbarUserNode} className="sui-TopbarUser">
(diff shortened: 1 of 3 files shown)


devictoribero:
which one do you mean @andresin87 ? any suggestion tho?

components/topbar/user/src/index.js
```diff
@@ -225,19 +223,22 @@ export default function TopbarUser({
...
+      {customContent}
+      {navCTA && !customContent && (
+        <div className="sui-TopbarUser-ctaButton">
+          <AtomButton
+            link
+            href={navCTA.url}
+            title={navCTA.text}
+            leftIcon={<navCTA.icon svgClass="sui-TopbarUser-ctaButtonIcon" />}
+            size="small"
+            type="primary"
```
jordevo, line 235: 🍆 I know it's unrelated to this `PR` but perhaps you could use size and type values as imported from `AtomButton`?

components/topbar/user/src/index.js
```diff
@@ -347,6 +348,11 @@ TopbarUser.propTypes = {
       })
     )
   }).isRequired,
+  /**
+   * Render custom content instead of a CTA.
+   * If a customContent is provided, it's rendered instead of a navCTA
+   */
+  customContent: PropTypes.node,
```
jordevo, line 355: 🍆 just wondering... would you see any benefit in using `children` instead of this `customContent` prop?

devictoribero: IMHO children should be used when you want to place something inside, it's a too strict API. @jordevo

components/topbar/user/src/index.js
devictoribero: I can, yes

components/topbar/user/src/index.js
jordevo: fair enough! 👍

PR #296, opened 2020-07-29
fix header color to white

paulusrex:
## Description
During the integration of TCF in IJ the background color of the modal header was the primary of IJ
Screenshot 2020-07-29 at 10 58 52

During the integration in CN/MN the background color of the first layer was grey
cn-desktop-firstLayer

## GOAL

Fix the color to white according UX specs

ij-desktop-secondLayer

cn-desktop-firstLayer-white


midudev:
cc/ @jcger

PR #303, opened 2020-08-07
fix(script/jsonLD): use dangerouslySetInnerHTML for avoid issues on b…

rmoralp:
Without `dangerouslySetInnerHTML` we detected that sui-bundler build it and replace spaces by html entities and Google Validation tool don't like it

### DIFF
--- a/components/script/jsonLD/src/index.js
+++ b/components/script/jsonLD/src/index.js
@@ -5,7 +5,10 @@ function ScriptJsonLD({json}) {
   return (
     <>
       {json && (
-        <script type="application/ld+json">{JSON.stringify(json)}</script>
+        <script
+          dangerouslySetInnerHTML={{__html: JSON.stringify(json)}}
+          type="application/ld+json"
+        />
       )}
     </>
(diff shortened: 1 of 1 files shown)


components/script/jsonLD/src/index.js
```diff
@@ -5,7 +5,10 @@ function ScriptJsonLD({json}) {
   return (
     <>
       {json && (
-        <script type="application/ld+json">{JSON.stringify(json)}</script>
+        <script
+          dangerouslySetInnerHTML={{__html: JSON.stringify(json)}}
```
nucliweb, line 9: We have **[htmlStringToReactElement](https://github.com/SUI-Components/sui/tree/master/packages/sui-js#string)**

```js
import {htmlStringToReactElement} from '@s-ui/js/lib/string'

htmlStringToReactElement('<p>No more dangerouslySetInnerHTML</p>')
```

rmoralp: it is not HTML string

PR #307, opened 2020-08-18
feat(form/builder): add numeric money field input rightAddon € icon

tomasmax:
Add right addon "€" icon to numeric money field input .


arnau-rius:
Since version 3. Of form vuilder you must add this type of customization via the children prop.

PR #331, opened 2020-11-05
feat(react/hooks): add a new useSwipe hook

ruben-martin-lozano:


danivm (approved):

desko27 (approved):

components/react/hooks/src/useSwipe/index.js
```diff
@@ -0,0 +1,80 @@
...
+
+export default function useSwipe(ref) {
+  const [start, setStart] = useState({x: 0, y: 0, time: 0})
+  const [capturing, setCapturing] = useState(false)
+  const [end, setEnd] = useState({x: 0, y: 0})
+  const [direction, setDirection] = useState(undefined)
+
+  const onStart = ({clientX: x, clientY: y}) => {
+    setCapturing(true)
+    setStart({x, y, time: +new Date()})
```
rmoralp, line 17: I think that if we use useReducer instead useState we could improve the number of renders in this hook. Given `onStart` updates two states,  useSwipe will re-render twice for each listener update.
Did you see some problem related to that in components that uses this hook?

components/react/hooks/src/useSwipe/index.js
```diff
@@ -0,0 +1,80 @@
...
+      if (capturing) {
+        current.addEventListener('pointermove', onMove)
+      } else {
+        current.removeEventListener('pointermove', onMove)
+
+        const {TIME} = THRESHOLDS
+        const deltaTime = +new Date() - start.time
+
+        if (deltaTime > TIME) {
+          setEnd()
```
rmoralp, line 35: `end` has an initial state `{x: 0, y: 0}`, I think that this line will set `end` value to undefined. Is it correct?

juanmendez90 (changes requested):
There is no easter egg you liar

components/react/hooks/src/useSwipe/index.js
```diff
@@ -0,0 +1,80 @@
...
+    const {current} = ref
+
+    if (current) {
+      if (capturing) {
+        current.addEventListener('pointermove', onMove)
+      } else {
+        current.removeEventListener('pointermove', onMove)
+
+        const {TIME} = THRESHOLDS
+        const deltaTime = +new Date() - start.time
```
rmoralp, line 32: I think that `Date.now()` is faster than `+new Date()` and easier to understand 
```suggestion
        const deltaTime = Date.now() - start.time
```

components/react/hooks/src/useSwipe/index.js
ruben-martin-lozano: Yes, I'll set the initial state to undefined since it will be more consistent :)

components/react/hooks/src/useSwipe/index.js
ruben-martin-lozano: It actually does not render twice: React automatically batches multiple changes in the state into once :)

rmoralp (approved):

juanmendez90 (approved):

nucliweb (approved):

PR #334, opened 2020-11-13
feat(react/hooks): add hooks for data fetching

andresz1:
## Hooks for data fetching
This PR adds hooks for fetching and updating asynchronous data in React easily ✨  . These hooks are inspired by [react-query](https://github.com/tannerlinsley/react-query). They have a similar API but without the caching feature.

This was made during a lab (ft @arnau-rius, @tomasmax, @javiauso and @igncalleja)

Main features:
- Abstraction over query `statuses` (e.g. `isLoading`, `isSuccess`, `isError`)
- Automatic query cancellation when component is unmounted
- Supports SSR using initial data
- Fetch data using a polling strategy out of the box
- Fetch data automatically when component mounts or just when is required
- Works great using DDD

### useQuery
The main purpose of this hook is to fetch data from a source.
```jsx
function Ads() {
   const {domain} = useContext(Context)
   const {isLoading, isError, data: ads, error} = useQuery(() => {
     return domain.get('get_ads_use_case').execute()
   })
 
   if (isLoading) {
     return <span>Loading...</span>
   }
 
   if (isError) {
     return <span>Error: {error.message}</span>
   }
 
   // also status === 'success', but "else" logic works, too
   return (
     <ul>
       {ads.map(ad => (
         <li key={ad.id}>{ad.title}</li>
       ))}
     </ul>
   )
 }
```

### useMutation
The main purpose of this hook is to create, update or remove data from a source.
```jsx
function Ad() {
   const {domain} = useContext(Context)
   const [createAd, {isLoading}] = useMutation(() => {
     return domain.get('create_ad_use_case').execute()
   }, {
     onSuccess: ad => {
       console.log(ad)
     }
   })

  const handleClick = () => {
    createAd()
  }

   return <button onClick={handleClick} isLoading={isLoading}>Create</button>
}
```


tonyghiani (commented):
HI @andresz1, I just saw this, I didn't see it was there as a draft!

First of all, good job on adopting the idea behind the react-query library, I'm a big fan🎉

Checking the implementation of this API, I see the similarities with the library , my main concern is about the features which makes react-query so awesome: cache and reactivity to queries update.

The cache management is probably one of the best features of react-query, and i have doubts if could be more beneficial to develop and maintain a so important feature, when we have the library already tested out-of-the-box. 

About the reactivity, at the current implementation state every hook call relies on its own state, meaning we won't have reactivity on query updates since there is not unique source-of-truth for the queries' result. I'm sure you thought about it for future iterations.

Concluding, I'm 100% in your team for adopting this management system for server-state, I just think could be faster to directly exploit the library instead reproducing a built in house version of it.

Let me know your thought, and still great job👏🏻👏🏻👏🏻

andresz1:
Hey @tonyghiani!

I'm huge fan of react-query too! glad you liked the idea. We did a PoC using it (we also tried the devtools) in a lab and it was pretty cool. Since we already have a caching layer using decorators (see [docs](https://github.com/SUI-Components/sui/tree/master/packages/sui-decorators#cache)) that works with memory (Client-Side) and [redis](https://redis.io/) (Server-Side), we thought that maybe we could start with a similar API to standardize, centralise and abstract features like: abstraction over query statuses, automatic query cancellation, retries, polling, among others which are currently repeated over and over again in our source code and increase our DX.

Nevertheless, I also agree that would be nice to just use react-query (since we don't need to maintain it, has the cache layer, devtool and more awesome stuff). Let's discuss how can we move this forward and check what to do with our current caching strategy.

PR #341, opened 2020-11-24
fix(form/builder): case sensitive error

arnau-rius:
case sensitive error


midudev (approved):

javiauso (approved):

Sabri1209 (approved):

arnau-rius:
I have renamed multipicker dir to multicheckbox because it is imposible to rename the dir using git... so the case sensitive import works..

davidmartin84 (approved):

quinwacca (approved):

PR #342, opened 2020-11-25
fix(image/slider): enable width and height attributes

rmoralp:
Add width and height attributes for slider images if image object has those keys.

### DIFF
--- a/components/image/slider/src/index.js
+++ b/components/image/slider/src/index.js
@@ -28,7 +28,16 @@ const defaultCounterPatternFactory = ({current, total}) => `${current}/${total}`
  */
 const getSlides = (currentSlide, images = [], linkFactory) => {
   return images.map((image, index) => {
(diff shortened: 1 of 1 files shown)


AgonisticKatai (approved):

ivanmlaborda (approved):

components/image/slider/src/index.js
```diff
@@ -39,6 +48,8 @@ const getSlides = (currentSlide, images = [], linkFactory) => {
         key={key}
         src={src}
         title={title}
+        height={height}
+        width={width}
```
nucliweb, line 52: 🍆  What happens if the height and/or width are `undefined`?

rmoralp: this attribute won't be printed

adriadoz (approved):

tomcask (approved):

midudev (approved):

PR #353, opened 2021-01-14
feat(form/captcha): prevent non rendering in multiples instances

AgonisticKatai:
### GOAL
Due to the need that we have in Motor to show in the DOM two instances of the FormCaptcha component, we must adapt the component to accept a unique identifier of the container where the Captcha will be injected, because otherwise, we are trying to inject two instances in the same node, which is not possible.

Captura de pantalla 2021-01-14 a las 17 37 33
Captura de pantalla 2021-01-14 a las 17 37 49

### DIFF
--- a/components/form/captcha/src/index.js
+++ b/components/form/captcha/src/index.js
@@ -17,6 +17,7 @@ const CAPTCHA_VERIFIER = () =>
   typeof window.grecaptcha.render !== 'undefined'
 
 const FormCaptcha = ({
+  containerId,
   siteKey,
   locale,
   onSubmit = () => {},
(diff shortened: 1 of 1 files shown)


jordevo (approved):

nucliweb (approved):

components/form/captcha/src/index.js
```diff
@@ -25,6 +26,9 @@ const FormCaptcha = ({
   const [showCaptcha, setShowCaptcha] = useState(false)
   const [captchaId, setCaptchaId] = useState(null)
 
+  // If we receive a unique identifier, we concatenate it with the default.
+  const captchaContainerId = [CAPTCHA_ID, containerId].filter(Boolean).join(' ')
```
jordevo, line 30: 🍆 why don't you just use one (the default one) or the other instead of joining them together?

raulmarindev (approved):

components/form/captcha/src/index.js
AgonisticKatai: Well, for nothing special, but as it refers to the class of the component, I wanted to keep it. 😛

rmoralp (approved):

components/form/captcha/src/index.js
jordevo: I wouldn't have used a blank space to join them... it's an id! ;)

PR #355, opened 2021-01-15
feat(map/basic): add pan and drag cb

andresz1:
This PR:
- Updates the map center when it's changed (only works if `isSetViewEnabled` prop is set to `true`)
- Adds `onMapDrag` prop


javiauso (approved):

desko27:
Hey @andresz1 since this component is very core here in FC I'd ask to let us test it before merging. Thanks, I'll keep you updated!

desko27 (approved):
Everything works fine now, thanks!!

PR #357, opened 2021-01-26
feat(tcf/ui): add testid in accept all button

paulusrex:
## Description
Add testid attribute for integration testing.

## Solves ticket
https://jira.scmspain.com/browse/PSP-3772
ISSUES CLOSED: PSP-3772


oscar-raig (approved):

alextremp (approved):
maybe it should also be done in the configure button and in the second layer accept button?

jcger (approved):

PR #375, opened 2021-03-10
feat(react/hooks): add setControlledState

andresin87:
## Description
The `useControlledState` hook has the purpose of unify the logic about combining the value and defaultValue behaviors of a component.

The hook provides a filtered value updater under the controlled mode in order to delegate responsibility of updating the value to its parent.

consumer can also check the mode which is working each moment and get the initial value any time.

## Usage
```js
const Component = ({value, initialValue, onClick, onReset}) => {
  const [
    innerValue, // state
    setInnerValue, // state setter
    isControlledValue, // {boolean} using controlled behavior or not
    initialInnerValue // first inner value state
  ] = useControlledState(value, initialValue)
  const onHandler = (handler, value) => () => {
    setInnerValue(value)
    typeof handler === 'function' && handler(value)
  }
  return (
    <FooBar
      value={innerValue}
      onClick={onHandler(onClick, innerValue)}
      onReset={onHandler(onReset, initialInnerValue)}
    />
  )
}
```

## Tasks

- [x] Add hook implementation
- [x] Add tests
- [x] Add demo
- [x] Add doc in Readme


tonyghiani (approved):

demo/react/hooks/demo/index.js
```diff
@@ -19,6 +19,7 @@ import UseScrollDemo from './UseScrollDemo'
 import UseStepsDemo from './UseStepsDemo'
 import UseSwipeDemo from './UseSwipeDemo'
 import UseToggleDemo from './UseToggleDemo'
+import UseControlledState from './UseControlledState'
```
tonyghiani, line 22: Just a small point, would be nice to keep the files naming consistent with the other hook demos :)

andresin87: done!

PR #384, opened 2021-04-15
feat(react/hooks): add new hook useDynamicHeightCollapsible

ferransimon:


components/react/hooks/README.md
```diff
@@ -451,3 +451,42 @@ const Component = ({value, initialValue, onClick, onReset}) => {
...
+### useDynamicHeightCollapsible
+
+> The `useDynamicHeightCollapsible` hook has the purpose of controlling the MoleculeCollapsible component when you need its height to be calculated dynamicaly.
+
+```js
+import {useRef} from 'react'
+
+const containerListRef = useRef()
+
+const getContentHeight = (
```
tonyghiani, line 464: Where are the properties for the `getContentHeight` function coming from? I see you are passing its reference to the hook, which does not provide those props

ferransimon: it is just an example, they are defined directly from the Component where the hook is used.

components/react/hooks/src/useDynamicHeightCollapsible/index.js
```diff
@@ -0,0 +1,17 @@
+import {useState, useCallback} from 'react'
+import useToggle from '../useToggle'
+
+export default function useDynamicHeightCollapsible(getContentHeighFn) {
+  const [isCollapsed, toggleCollapsible] = useToggle(true)
+  const [contentHeight, setContentHeight] = useState(null)
+  const heightRecalculator = useCallback(() => {
```
tonyghiani, line 7: I think in this case adding the useCallback hook could be a deoptimization because of these reasons:
- The dependency is a function, which most of the times will be declared inline or will need to be memorized by the parent component, which will include more complexity.
- Memoizing this function makes sense if the main purpose is to pass its reference to an atomic component event or a memorized component, like a button for example. Is this the main scope of this function?
I think would be better to remove it until the use case of this function is not clear in different cases, we can always refactor it and add the memoization.

components/react/hooks/src/useDynamicHeightCollapsible/index.js
```diff
@@ -0,0 +1,17 @@
...
+    const newContentHeight = getContentHeighFn()
+    setContentHeight(newContentHeight)
+  }, [getContentHeighFn])
+
+  return [
+    isCollapsed,
+    contentHeight,
+    {toggle: () => toggleCollapsible(), heightRecalculator}
```
tonyghiani, line 15: `toggle` will always be a new function if declared like this, use the already memorized function from useToggle:
```jsx
const [isCollapsed, toggle] = useToggle(true)
...
{toggle, heightRecalculator}
```

components/react/hooks/src/useDynamicHeightCollapsible/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+export default function useDynamicHeightCollapsible(getContentHeighFn) {
+  const [isCollapsed, toggleCollapsible] = useToggle(true)
+  const [contentHeight, setContentHeight] = useState(null)
+  const heightRecalculator = useCallback(() => {
+    const newContentHeight = getContentHeighFn()
+    setContentHeight(newContentHeight)
+  }, [getContentHeighFn])
+
+  return [
```
tonyghiani, line 12: This hook will return more than a state and more than a handler. In this case, it's better to return an object of named properties, since the order of the array make the restructuring error prone. I'd suggest something like:
```js
return {
  isCollapsed,
  height,
  toggle,
  heightRecalculator
}
```

components/react/hooks/src/useDynamicHeightCollapsible/index.js
ferransimon: Yeah, but I need the toggle to not allow passing an initialValue since the SUI-MoleculeCollapsible does not allow to specify an initial value for the open/close state

components/react/hooks/README.md
ferransimon: You are right I need at least the ref current to be passed to the function

components/react/hooks/src/useDynamicHeightCollapsible/index.js
tonyghiani: Mmm what initial value? the toggle function accepts only  a boolean as argument, otherwise it'll be ignored. It should be safe if this was the problem, otherwise I need to better understand the specific case you are talking about 😃

components/react/hooks/src/useDynamicHeightCollapsible/index.js
ferransimon: done!

components/react/hooks/src/useDynamicHeightCollapsible/index.js
ferransimon: done!

components/react/hooks/src/useDynamicHeightCollapsible/index.js
ferransimon: done!

components/react/hooks/README.md
ferransimon: removed the function from the hook parameters so now it is defined inside the hook so this hook is expecting a children array to calculate its content height by iterating over the number of visible items that the user of the hook has specified. So if the number of visibleItems changes the useEffect will execute the heightRecalculator with the reference

PR #397, opened 2021-05-12
refactor(react/hooks): unnecesary hooks and reusability

andresin87:
- Removes useSetState (unnecesary)
- Refactor of useMount (removes unused arguments)
- Refactor and use useMount on useControlledState
- Refactor useCopyToClipboard (pase useSetState usage to useState)


components/react/hooks/src/useCopyToClipboard/index.js
```diff
@@ -1,15 +1,12 @@
...
 import useMountedState from '../useMountedState'
-import useSetState from '../useSetState'
 
 export default function useCopyToClipboard() {
   const isMounted = useMountedState()
-  const [state, setState] = useSetState({
-    value: undefined,
-    error: undefined
-  })
+  const [value, setValue] = useState()
```
jcger, line 8: as you are setting these values always as one, what about using something like `const [{value, error}, setState]` so you can `setState({value, error})` afterwards without triggering 2 renders?

andresin87: done

components/react/hooks/src/useMountedState/index.js
```diff
@@ -1,6 +1,6 @@
 import {useCallback, useEffect, useRef} from 'react'
 
-export default function useMountedState(value, delay) {
+export default function useMountedState() {
```
tonyghiani, line 3: I have doubts about this hook, why not returning this one as a boolean directly more than a getter function?

components/react/hooks/src/useControlledState/index.js
```diff
@@ -1,21 +1,20 @@
...
     controlledValue === undefined ? defaultValue : controlledValue
   )
   const [value, setValue] = useState(initialValue)
-  useEffect(() => {
-    if (isFirst.current) {
-      isFirst.current = false
-    } else {
+  const isMounted = useMountedState()
+  useLayoutEffect(() => {
+    if (isMounted()) {
```
rmoralp, line 14: I think that it changes behaviour. In changed code isFirst was set to true, then moved to false after "mount" and then `setValue`
Isn't it?
```suggestion
    if (isMounted() === false) {
```

components/react/hooks/src/useControlledState/index.js
```diff
@@ -1,21 +1,20 @@
...
   const [initialValue] = useState(
     controlledValue === undefined ? defaultValue : controlledValue
   )
   const [value, setValue] = useState(initialValue)
-  useEffect(() => {
-    if (isFirst.current) {
-      isFirst.current = false
-    } else {
+  const isMounted = useMountedState()
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
andresin87: isMount uses the useEffect on it and use the useRef for the mounted status acknowledge. In order not to depend on the useEffect declaration position preventing posible bugs we might use This useLayoutEffect.

[@kentcdodds](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect) (last example –One special case
–)

components/react/hooks/src/useControlledState/index.js
andresin87: Its a semantic difference. isFirst means isNotMounted

components/react/hooks/src/useMountedState/index.js
andresin87: It is using the reference for the behavior and no-useEffect should depend on the state of the hooks inner value, so we only should return a memoized callback targetting to the value (no other component observers will be refreshed when its inner status is updated)

take a look to the [https://github.com/streamich/react-use/blob/master/src/useMountedState.ts](https://github.com/streamich/react-use/blob/master/src/useMountedState.ts) code example

tonyghiani: Sure i understand, what i mean is that even returning directly the boolean value and not a getter it shouldn't notify and rerender any component observing this value

tonyghiani (approved):

jcger (approved):

PR #401, opened 2021-05-27
feat(image/slider): be able to send custom images to react-slidy

AgonisticKatai:
Feature to be able to render a component as a card inside the image gallery itself.
This is already supported by **react-slidy**, but as we have implemented the component, we only allow to display images through an img tag.

### WITH IMAGE TYPE SRC
Captura de pantalla 2021-05-27 a las 7 18 53

### WITH IMAGE TYPE CARD
Captura de pantalla 2021-05-27 a las 7 19 06

### DIFF
--- a/components/image/slider/src/index.js
+++ b/components/image/slider/src/index.js
@@ -31,7 +31,7 @@ const defaultCounterPatternFactory = ({current, total}) => `${current}/${total}`
  * @return {Array} List of img or node elements.
  */
 const getSlides = (currentSlide, content = [], linkFactory) => {
-  return content.map((content, index) => {
+  return content.map((contentItem, index) => {
     const {
       alt,
       height,
@@ -42,7 +42,7 @@ const getSlides = (currentSlide, content = [], linkFactory) => {
       title,
       type = IMAGE_SLIDER_CONTENT_TYPES.IMAGE,
       width
-    } = content
+    } = contentItem
 
     const key = imageKey ? imageKey + index : index
     const children =
@@ -58,7 +58,7 @@ const getSlides = (currentSlide, content = [], linkFactory) => {
           width={width}
         />
       ) : (
-        content
+        contentItem
       )
 
     return link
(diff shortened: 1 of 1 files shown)


nucliweb (approved):

components/image/slider/src/index.js
```diff
@@ -9,6 +9,10 @@ export const IMAGE_SLIDER_COUNTER_POSITIONS = {
   BOTTOM_LEFT: 'bottomLeft',
   BOTTOM_RIGHT: 'bottomRight'
 }
+export const IMAGE_SLIDER_IMAGE_TYPES = {
+  SRC: 'src',
+  CARD: 'card'
```
jordevo, line 14: IMHO it'd make more sense to me to name the default type `img` as it ends up rendering an image, and whatever you can pass it on, the other type, could be just a `node`... what do you think?

```suggestion
  IMG: 'img',
  NODE: 'node'
```

components/image/slider/src/index.js
```diff
@@ -36,31 +40,36 @@ const getSlides = (currentSlide, images = [], linkFactory) => {
       src,
       target = TARGET_BLANK,
       title,
+      type = IMAGE_SLIDER_IMAGE_TYPES.SRC,
       width
     } = image
```
andresin87, line 45: Replace image as content to be more specific

andresin87:
Its a little bit crazy to have a ImageSlider component which renders elements that are not images.

components/image/slider/src/index.js
andresin87:
```js
const getSlides = (currentSlide, content = [], linkFactory) => {
///...
```

andresin87: The JSDOC is not properly written

components/image/slider/src/index.js
```diff
@@ -9,6 +9,10 @@ export const IMAGE_SLIDER_COUNTER_POSITIONS = {
   BOTTOM_LEFT: 'bottomLeft',
   BOTTOM_RIGHT: 'bottomRight'
 }
+export const IMAGE_SLIDER_IMAGE_TYPES = {
```
andresin87, line 12: IMAGE_SLIDER_CONTENT_TYPES

jordevo (approved):

kikoruiz (approved):

AgonisticKatai:
@andresin87 I agree, but I understand that it is a slider and maybe it should not be coupled only to images.
We could even consider the possibility of changing it from the context images

andresin87 (approved):

PR #421, opened 2021-08-25
Scmi 61826 add extra component to pta form

fredyzz:
### ✅ Types of changes

- New feature (non-breaking change which adds functionality)

### 📝 Description, Motivation and Context

Add button display to multipicker field

### 📸 Screenshots - Animations
Captura de pantalla 2021-08-25 a las 16 21 57

### DIFF
--- a/components/form/builder/src/MultiButton/index.js
+++ b/components/form/builder/src/MultiButton/index.js
@@ -1,0 +1,115 @@
+import {isValidElement, memo} from 'react'
+import PropTypes from 'prop-types'
+import AtomTag from '@s-ui/react-atom-tag'
+import MoleculeField from '@s-ui/react-molecule-field'
+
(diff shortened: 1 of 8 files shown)


components/form/builder/src/MultiButton/index.js
```diff
@@ -0,0 +1,115 @@
...
+  const onClickCallback = (id, checked) => {
+    const targetId = id
+
+    let newValue = []
+
+    if (checked) {
+      newValue = multiButton.value.filter(id => id !== targetId)
+    } else {
+      newValue = [...(multiButton.value || []), targetId]
+    }
```
klaufel, line 30: In this way we avoid mutations

```suggestion
    const newValue = checked
      ? multiButton.value.filter(id => id !== targetId)
      : [...(multiButton.value || []), targetId]
```

PR #422, opened 2021-08-30
feat(form/builder): add new SUPERSET and NSUPERSET operators

oriolpuig:
### ✅ Types of changes

- New feature (non-breaking change which adds functionality)

### 📝 Description, Motivation and Context

Add two new operators: 
- SUPERSET: Check if a list of rule values contains the selected field values. 
- NSUPERSET: inverse of SUPERSET.

### 🔢 Example

```
// SUPERSET ✅
const rules = [1, 2]
const selected = [1, 2, 3]
result => true

// SUPERSET ❌
const rules = [1, 3]
const selected = [1, 2]
result => false
```


demo/form/builder/demo/formList.js
```diff
@@ -1179,9 +1179,31 @@ export const json = {
           {value: '96', text: 'Sólo chicos'},
           {value: '97', text: 'Sólo no fumadores'}
         ]
+      },
+      {
+        id: 'priceDaily',
```
Alverd04, line 1184: order properties

Alverd04 (changes requested):

adriadoz (approved):

nistadev (approved):

PR #423, opened 2021-09-01
chore(Root): Migrate to sui-studio-10

midudev:
Use new sui-studio@10


.github/workflows/node.js.yml
```diff
@@ -15,62 +15,20 @@ jobs:
...
-    - name: node_modules cache
-      id: cache
-      uses: actions/cache@v2
-      with:
-        path: |
-          **/node_modules
-        key: ${{ github.run_number }}
-    - name: Build dependencies
-      run: npm run build
+        npm install --legacy-peer-deps --no-audit --no-fund --no-optional
```
andresz1, line 29: Maybe adding cache here would be nice ? Maybe using something like this [action](https://github.com/bahmutov/npm-install)

andresz1 (approved):

nucliweb (approved):

.github/workflows/node.js.yml
midudev: I've removed cache as it's not working as expected with workspaces as they use symbolic links for dependencies that are already available in the workspace.

xaviermariaconejo (approved):

andresin87 (approved):

nistadev (approved):

quinwacca (approved):

adriadoz (approved):

PR #424, opened 2021-09-02
feat(form/builder): add hidden operator and fix default case typo

jmanrumartinez:
Adds **HIDDEN** operator and fixes a typo

**How to use**
As a normal conditional, you should declare which field is checked.

```
{
      when: [
        {
          operator: 'EQUALS',
          id: FIELD_KEYS.WANT_TO_ADD_SECONDARY_PHONE,
          value: ['false']
        },
        {
          operator: 'NIN',
          id: FIELD_KEYS.MAIN_PHONE,
          value: ['']
        },
        {
          operator: 'HIDDEN',
          id: FIELD_KEYS.SECONDARY_PHONE
        }
      ],
      then: {
        data: {
          hidden: false
        }
      }
}
```

The rule will trigger if the `WANT_TO_ADD_SECONDARY_PHONE` value is `false`, there're value in `MAIN_PHONE` and the `SECONDARY_PHONE` is hidden. 

**Why it's needed and when to use**
In our case, we only want to show the wantToAddSecondaryPhone field if the `secondaryPhone` field is hidden, because there're cases like the `secondaryPhone` field is saved in localStorage and initially appears the two fields (main phone and secondary phone)

Captura de pantalla 2021-09-02 a las 11 25 22

So without the hidden condition will trigger because `WANT_TO_ADD_SECONDARY_PHONE` is initialized with false value, and there's value in the `MAIN_PHONE` (it will appear the `WANT_TO_ADD_SECONDARY_PHONE` field if the user rewrites the phone)


components/form/builder/src/reducer/operators.js
```diff
@@ -104,5 +105,9 @@ export const operators = {
     const isEmpty = !fieldValue.length
     const hasNotExpectedValues = values.some(i => !fieldValue.includes(i))
     return isEmpty || hasNotExpectedValues
+  },
+  [HIDDEN]: (id, fields) => {
+    const {hidden} = pickFieldById(fields, id)
+    return hidden === true
```
jmanrumartinez, line 111: Do you prefer to return hidden directly?

arnau-rius: yes

arnau-rius: If you want to make sure it is a boolean you can cast it using `Boolean(hidden)` or `!!hidden`

components/form/builder/src/reducer/rules.js
```diff
@@ -105,8 +107,11 @@ export const shouldApplyRule = (fields, changeField, locale) => when => {
       case NSUPERSET:
         isValid = operators.NSUPERSET(rule.id, rule.value, fields, locale)
         break
+      case HIDDEN:
+        isValid = operators.HIDDEN(rule.id, fields)
+        break
       default:
-        console.warn(`Unkown operator ${rule.operator}`)
+        console.warn(`Unknown operator ${rule.operator}`)
```
arnau-rius, line 114: ✅ speller

arnau-rius (approved):

components/form/builder/src/reducer/operators.js
andresz1: ⬆️ agree

andresz1:
IMO documentation + example would be awesome and why is the operator needed!

tomasmax (approved):

jmanrumartinez:
> IMO documentation + example would be awesome and why is the operator needed!

Updated PR with some explanation and updated the docs [here](https://github.mpi-internal.com/scmspain/all--lib-form-builder-docs/pull/13)

andresz1 (approved):

PR #428, opened 2021-09-22
Feature/topbar user menu events

AgonisticKatai:
### GOAL
Be able to send onClick menu event to parent

### DIFF
--- a/components/topbar/user/src/index.js
+++ b/components/topbar/user/src/index.js
@@ -156,9 +156,11 @@ export default function TopbarUser({
    * Render main navigation function.
    */
   const _renderNavMain = isToggleHidden => (
-    {icon, label: text, menu, arrowButtonIcon},
+    {icon, label: text, menu, arrowButtonIcon, onClick},
     index
   ) => {
+    const handleToggleMenu = () => onClick()
+
     return (
       <DropdownBasic
         key={index}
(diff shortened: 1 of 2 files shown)


components/topbar/user/src/index.js
```diff
@@ -156,16 +156,19 @@ export default function TopbarUser({
    * Render main navigation function.
    */
   const _renderNavMain = isToggleHidden => (
-    {icon, label: text, menu, arrowButtonIcon},
+    {icon, label: text, menu, arrowButtonIcon, onClick},
```
jordevo, line 159: wouldn't this serve the same purpose and save you from declaring a new variable (L162)?

```suggestion
    {icon, label: text, menu, arrowButtonIcon, onClick: handleToggleMenu},
```

jordevo (approved):

components/topbar/user/src/index.js
AgonisticKatai: I think it is interesting to leave the function ready in case in the future you want to do some previous action from the component itself before executing the onClick method, isn't it?

jordevo: fair enough... don't think it's a big deal performance wise but I always try to keep those "just in case" lines of code off the code, when possible 😉

AgonisticKatai: It's just one line of code 😛

kikoruiz (approved):

PR #434, opened 2021-11-03
feat(components/image/slider): use AtomImage instead default image tag

AgonisticKatai:
### GOAL

- [x] Use AtomImage instead native image tag
- [x] Be able to receive an array of image source to improve performance.

### DIFF
--- a/components/image/slider/package.json
+++ b/components/image/slider/package.json
@@ -12,9 +12,10 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "babel-runtime": "6.26.0",
     "@s-ui/component-dependencies": "1",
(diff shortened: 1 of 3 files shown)


jordevo (approved):

kikoruiz (approved):

components/form/builder/src/Standard/Fields/Numeric/index.js
```diff
@@ -45,6 +45,7 @@ const NumericField = ({
 }
 
 NumericField.propTypes = {
+  onBlur: PropTypes.func,
```
AgonisticKatai, line 48: 👀. Fix linter error

PR #438, opened 2021-11-11
Feat Form Builder Multipicker Dropdown Close Icon

jordevo:
A Close Icon was missing for `multipicker` fields with `dropdown` display.

image


AgonisticKatai (approved):

nucliweb (approved):

javiauso (approved):

oriolpuig (approved):

davidmartin84 (approved):

components/form/builder/src/Icons/IconCloseTag.js
```diff
@@ -0,0 +1,17 @@
+import {memo} from 'react'
+import PropTypes from 'prop-types'
+
+const SVG = ({className}) => (
+  <svg viewBox="0 0 24 24">
+    <path
+      id="a"
```
arnau-rius, line 7: maybe we should use a more generic id, to avoid collisions with other ids

jordevo: you're right! that's what copying and pasting from one of our icons gets us haha good catch!

jordevo: and actually, I don't even think an `id` is necessary here

kikoruiz (approved):

PR #439, opened 2021-11-23
Feat Checkbox nodeLabel prop

jordevo:
So far, we're not able to use a `node` as a label for a `checkbox` display on a `picker` type of field.

This *PR* addresses the issue by adding `nodeLabel` when building the `checkbox` component props.

Simple example:

image

### DIFF
--- a/components/form/builder/demo/formPTACar.js
+++ b/components/form/builder/demo/formPTACar.js
@@ -1519,7 +1519,9 @@ export const formPTACar = {
           {
             id: 'terms',
             type: 'picker',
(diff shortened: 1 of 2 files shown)


davidmartin84 (approved):

jordevo:
We're finally not going for this approach since it _breaks_ the _cross_ philosophy of the Form Builder API, being `nodeLabel` a specific React prop that would not be used on other platforms.

arnau-rius:
You can use the children: renderer prop in order to customize any prop that is passed to any field: https://github.com/SUI-Components/adevinta-spain-components/blob/master/components/form/builder/src/index.js#L35

PR #441, opened 2021-12-09
feat(components/form/builder): add applyEmptyFields prop

javiauso:
add applyEmptyFields prop

### DIFF
--- a/components/form/builder/src/index.js
+++ b/components/form/builder/src/index.js
@@ -18,7 +18,7 @@ import AtomSpinner, {AtomSpinnerTypes} from '@s-ui/react-atom-spinner'
 import {getUpdatedFormState} from './mapper/formState'
 
 const FormBuilder = ({
-  applyEmptyFields,
+  forceRulesOnAllFields,
   json,
   initFields = {},
   onChange = () => {},
@@ -113,7 +113,7 @@ const FormBuilder = ({
       .reduce(async (previousPromise, fieldId) => {
         const previousFields = await previousPromise
         if (
-          !applyEmptyFields &&
+          !forceRulesOnAllFields &&
           (!initFields[fieldId] || initFields[fieldId] === '')
         ) {
           return previousFields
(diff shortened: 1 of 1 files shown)


Sabri1209 (approved):

andresz1 (approved):

components/form/builder/src/index.js
```diff
@@ -178,20 +183,6 @@ FormBuilder.propTypes = {
   children: PropTypes.func
 }
 
-FormBuilder.defaultProps = {
```
jordevo, line 181: 👏

components/form/builder/src/index.js
```diff
@@ -18,21 +18,22 @@ import AtomSpinner, {AtomSpinnerTypes} from '@s-ui/react-atom-spinner'
 import {getUpdatedFormState} from './mapper/formState'
 
 const FormBuilder = ({
+  applyEmptyFields,
```
jordevo, line 21: what is the new prop about? by its name, I'm not sure it's self-explanatory. Also, should it default to `false`?

tomasmax (approved):

jordevo (approved):

components/form/builder/src/index.js
javiauso: This prop apply form builder logic like apply rules and set initialFields values for empty fields, for example, if has defined `{categoryId: '1', subCategoryId: ''}` without `applyEmptyFields` prop, will only apply rules and form logic for categoryId. The name probably could be improved, any suggestion 😄?

components/form/builder/src/index.js
```diff
@@ -18,21 +18,22 @@ import AtomSpinner, {AtomSpinnerTypes} from '@s-ui/react-atom-spinner'
...
   json,
-  initFields,
-  onChange,
-  onFocus,
-  onBlur,
-  responseInterceptor,
-  requestInterceptor,
+  initFields = {},
+  onChange = () => {},
```
andresin87, line 24: Functions should not be empty (Code Smell –Critical)
https://rules.sonarsource.com/javascript/RSPEC-1186

components/form/builder/src/index.js
javiauso: By default is falsy to not break actual behaviour

jordevo: my 2 cents if I may 😉

`shouldAlwaysApplyRules`
`forceRulesOnAllFields`

I miss the `rules` part on `applyEmptyFields` and I think it's key to understand the purpose of the prop...

kikoruiz (approved):

components/form/builder/src/index.js
javiauso: I like `forceRulesOnAllFields`, I think it's clearer, thank you!

PR #445, opened 2021-12-27
feat(components/react/hooks): add new hook useInterval

oriolpuig:
- Add new useInterval hook based on [react-use example](https://github.com/streamich/react-use/blob/master/src/useInterval.ts)
- change exports to be ESModules friendly
- Fix phoenix npm script


midudev (approved):

andresin87:
a demo may help:
https://github.com/SUI-Components/adevinta-spain-components/tree/master/components/react/hooks/demo

PR #452, opened 2022-01-24
feat(components/dropdown/basic): add variable to modify font-weight

xmurcia:
Use new variable recently created in sui-theme as `inherit` by default. In motor we need to overwrite it to semibold

### DIFF
--- a/components/dropdown/basic/src/index.scss
+++ b/components/dropdown/basic/src/index.scss
@@ -63,6 +63,7 @@
       }
       align-items: center;
       display: flex;
+      font-weight: $fw-dropdown-basic-button;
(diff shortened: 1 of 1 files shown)


components/dropdown/basic/src/index.scss
```diff
@@ -63,6 +63,7 @@
       }
       align-items: center;
       display: flex;
+      font-weight: $fw-dropdown-basic-button;
```
jordevo, line 66: shouldn't this be set at `&-button` level?

quinwacca (approved):

tomcask (approved):

PR #455, opened 2022-01-31
feat(components/topbar/user): Add a scss token to topbar/user to make the icon color customizable from the app theme

oegea:
## Changes

A scss token has been added to the topbar/user component, in order to make customizable the toggle icon color.
Before this change, the topbar icon color was always accent. Now it's accent by default, but can be changed from the app theme.

## Context

coches.net rebranding experiment

## Change type

[x] New feature, not breaking change

### DIFF
--- a/components/topbar/user/src/index.scss
+++ b/components/topbar/user/src/index.scss
@@ -6,17 +6,18 @@
 @import '~@s-ui/react-atom-button/lib/index';
 
 // Topbar user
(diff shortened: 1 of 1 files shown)


oegea:
Btw, note that tokens have been alphabetically re-ordered. But actually just one new token has been added.

kikoruiz (approved):

PR #466, opened 2022-02-16
feat(components/ad/smartbanner): Add the possibility of defining a custom button element in smartbanner

oegea:
# Why 

In coches.net, we need to be able to pass a custom element to be used to render the install button of the smartbanner. In a similar way we can now pass a custom element to represent the close button.

Thanks to this change, we will be able to pass a standard AtomButton component that will have the same behaviour and styles than the rest of buttons in the web app.

# What this change does?

This change adds a new prop to the smartbanner component. If the prop is not defined, the component will behave as always with no changes, if the prop is present, its value will be used to render the install button.

# Context

Coches.net rebranding (UX final adjustments).

### DIFF
--- a/components/ad/smartbanner/src/index.js
+++ b/components/ad/smartbanner/src/index.js
@@ -1,9 +1,10 @@
 import PropTypes from 'prop-types'
 import IconCloseDefault from '@s-ui/react-icons/lib/X'
 import cx from 'classnames'
-import RatingStar from './RatingStar'
+import RatingStar from './RatingStar/index.js'
 
 function AdSmartbanner({
+  button: Button,
   buttonText,
   customRatingIcons,
   icon: IconClose = IconCloseDefault,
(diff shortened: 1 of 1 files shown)


andresin87:
You can remove the rendered ternary if you define a default button prop spreaded component in default defined Button redefined prop. Easy to read.

oegea:
@andresin87 Done!

components/ad/smartbanner/src/InstallButton/index.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types'
+
+const InstallButton = ({children, onClick}) => {
+  const baseClass = 'sui-AdSmartbanner'
```
arnau-rius, line 4: is it possible to import it from src/index.js file?

components/ad/smartbanner/src/index.js
```diff
@@ -49,14 +51,13 @@ function AdSmartbanner({
...
-      <button className={`${baseClass}-buttonInstall`} onClick={onClick}>
-        {buttonText}
-      </button>
+      <Button onClick={onClick}>{buttonText}</Button>
     </div>
   )
 }
 
 AdSmartbanner.propTypes = {
+  button: PropTypes.func,
```
arnau-rius, line 60: It is better to use `PropTypes.node` since it accepts not only functions but: numbers, strings, elements or an array

arnau-rius: Check it out: https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

javiauso (approved):

midudev (approved):
👍 but look to @arnau-rius comments.

components/ad/smartbanner/src/index.js
```diff
@@ -1,12 +1,14 @@
-import PropTypes from 'prop-types'
-import IconCloseDefault from '@s-ui/react-icons/lib/X'
 import cx from 'classnames'
-import RatingStar from './RatingStar'
+import DefaultCloseIcon from '@s-ui/react-icons/lib/X'
```
andresin87, line 2: 👀

PR #478, opened 2022-04-11
Feat/use callback ref

andresz1:
This PR add `useCallbackRef` hook which persists a function between renders but keeps it up-to-date if it changes.


andresin87:
useIsomorphicLayoutEffect is not in readme: https://github.com/streamich/react-use/blob/master/docs/useIsomorphicLayoutEffect.md

useIsomorphicLayoutEffect and useCallbackRef does not have a demo

tonyghiani (commented):
I see why you want to use the latest ref pattern here to memoize the callback even if it changes, but I've got some doubt about it:
- In the React community, the useCallbackRef custom hook is commonly used to apply a callback each time a ref updates or to merge refs, something like
```js
const ref = useCallbackRef(value, () => console.log('The ref changed!'))
```
So I don't know if this could be really the right name for the functionality you want.
- Then, my biggest concern is about the benefit of this implementation, isn't it adding an overhead to the rendering lifecycle, once what you really want is to memoize only when nothing changes?

Maybe I'm missing something, would be nice if you can give us an example where you measured it is beneficial its usage over the native useCallback() 😄

andresz1:
Some updates based on the reviews:
- Added a `useIsomorphicLayoutEffect` to README
- Added a link to the pattern used in the `useCallbackRef` hook. Thanks @tonyghiani 🔝 
- I didn't add examples because I think the hooks are no that UI related
- Fix tests

A better example:
The `useEventListener` hook is pretty useful.

```js
function useEventListener(eventName, handler, element = window) {
  // ------------------- The Latest Ref Pattern
  // Create a ref that stores handler
  const savedHandler = useRef();

  // Update ref.current value if handler changes.
  // This allows our effect below to always get latest handler ...
  // ... without us needing to pass it in effect deps array ...
  // ... and potentially cause effect to re-run every render.
  useEffect(() => {
    savedHandler.current = handler;
  }, [handler]);
  // --------------------

  useEffect(
    () => {
      // Make sure element supports addEventListener
      // On
      const isSupported = element && element.addEventListener;
      if (!isSupported) return;

      // Create event listener that calls handler function stored in ref
      const eventListener = (event) => savedHandler.current(event);

      // Add event listener
      element.addEventListener(eventName, eventListener);

      // Remove event listener on cleanup
      return () => {
        element.removeEventListener(eventName, eventListener);
      };
    },
    [eventName, element] // Re-run if eventName or element changes
  );
}
```

We use a handler to subscribe to a given event and we do not need to care about of adding and removing the event. Normally the handler function is anonymous one.

```js
function Test() {
  useEventListener('keydown', () => {
    console.log('keydown')
  })
}
```

If we don't use [The Latest Ref Pattern](https://epicreact.dev/the-latest-ref-pattern-in-react/) here (see code above) and we add the `handler` as a dependency to the `useEffect` it would be executed during each render removing and adding the event again (waste of time). This pattern persists the function between renders (in this case the `handler`) but keeps it up-to-date if it changes.

Using the `useCallbackRef` the previous hook could be refactored as follows:

```js
function useEventListener(eventName, handler: handlerProp, element = window) {
  // ------------------- The Latest Ref Pattern
  const handler = useCallbackRef(handlerProp);

  useEffect(
    () => {
      // Make sure element supports addEventListener
      // On
      const isSupported = element && element.addEventListener;
      if (!isSupported) return;

      // Create event listener that calls handler function stored in ref
      const eventListener = (event) => handler(event);

      // Add event listener
      element.addEventListener(eventName, eventListener);

      // Remove event listener on cleanup
      return () => {
        element.removeEventListener(eventName, eventListener);
      };
    },
    [eventName, element, handler] // Re-run if eventName or element changes
  );
}
```

javiauso (approved):

tonyghiani (approved):

rmoralp (approved):

tomasmax (approved):

jordevo (approved):

PR #485, opened 2022-05-11
feat(components/form/builder): add autocompleted field info on changing values

quinwacca:
## form/builder

Add autocompleted field info on changing values.
When `onChange` receives the next field value, `handlerChange` will be call with `true` value for `autocompletedField`, to add that info on `onChange`call.

It's not a breaking change. Just adds info.

We (motor) need to know if a field has been autocomplete, and which one.

### DIFF
--- a/components/form/builder/src/index.js
+++ b/components/form/builder/src/index.js
@@ -53,52 +53,56 @@ const FormBuilder = ({
     return nextFields
   }
 
-  const handlerChange = useCallback(async (id, value, extraParams) => {
-    const reducerWithRules = reducer(
-      rules,
-      formID,
-      baseAPIURL,
-      responseInterceptor,
-      requestInterceptor,
-      locale,
-      extraParams
-    )
-    const timerShowSpinner = setTimeout(
-      () => setStateShowSpinner(true),
-      FormBuilder.USER_MINIMAL_DELAY
-    )
-    const transformedValue = transformations(
-      id,
-      value,
-      pickFieldById(fields, id)
-    )
+  const handlerChange = useCallback(
+    async (id, value, extraParams, autocompletedField = false) => {
+      const reducerWithRules = reducer(
+        rules,
+        formID,
+        baseAPIURL,
+        responseInterceptor,
+        requestInterceptor,
+        locale,
+        extraParams
+      )
+      const timerShowSpinner = setTimeout(
+        () => setStateShowSpinner(true),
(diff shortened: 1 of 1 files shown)


components/form/builder/src/index.js
```diff
@@ -53,52 +53,56 @@ const FormBuilder = ({
...
-      () => setStateShowSpinner(true),
-      FormBuilder.USER_MINIMAL_DELAY
-    )
-    const transformedValue = transformations(
-      id,
-      value,
-      pickFieldById(fields, id)
-    )
+  const handlerChange = useCallback(
+    async (id, value, extraParams, autocompletedField = false) => {
```
quinwacca, line 57: There's a lot of lines change because the new tabulation, so I'm pointing the real changes.

Here, `onChange` receives a new param `autocompletedField`, `false` by default.

components/form/builder/src/index.js
```diff
@@ -53,52 +53,56 @@ const FormBuilder = ({
...
-        await handlerChange(key, value)
+      clearTimeout(timerShowSpinner)
+      setStateFields(nextStateFields)
+      const nextStateFieldsObject = useNativeFieldType
+        ? fieldsToObjectNativeTypes(nextStateFields)
+        : fieldsToObject(nextStateFields)
+
+      const fieldsToUpdate = await onChange({
+        ...nextStateFieldsObject,
+        __AUTOCOMPLETED_FIELD__: Boolean(autocompletedField),
```
quinwacca, line 93: Here we're adding more info in the `onChange` call.

components/form/builder/src/index.js
```diff
@@ -53,52 +53,56 @@ const FormBuilder = ({
...
+        __FIELD_CHANGED__: id,
+        __FORM_STATE__: getUpdatedFormState(json.form, nextStateFields)
       })
 
-    setStateShowSpinner(false)
+      if (typeof fieldsToUpdate === 'object' && fieldsToUpdate !== null) {
+        const autocompletedFields = true
+        Object.entries(fieldsToUpdate).forEach(async ([key, value]) => {
+          await handlerChange(key, value, undefined, autocompletedFields)
+        })
```
quinwacca, line 102: If exists a response on `onChange`, we can assume that a field will be autocomplete, so we call `handlerChange` with that param with a `true` value.

javiauso (approved):

PR #491, opened 2022-05-23
feat(components/form/builder): Create new range field type with a def…

Sabri1209:
**form/builder**
Create a new range field type with a default display being text range.
It has been created based on this [doc](https://docs.mpi-internal.com/scmspain/all--lib-form-builder-docs/form-specification/field/range/)
Screenshot 2022-05-23 at 15 56 41

**Screenshot**
image

### DIFF
--- a/components/form/builder/src/Range/Default/index.js
+++ b/components/form/builder/src/Range/Default/index.js
@@ -18,7 +18,6 @@ const splitInputValues = inputValues =>
 
 const DefaultRange = ({
   range,
-  tabIndex,
   onChange,
   onFocus,
   onBlur,
@@ -50,11 +49,10 @@ const DefaultRange = ({
   const onBlurCallback = ev => onBlur(ev.target.id)
   const onFocusCallback = ev => onFocus(ev.target.id)
 
-  const {datalist = []} = range
-
   const rangeProps = {
     id: range.id,
     display: range.display,
+    value: range.value,
     onChange: onChangeCallback,
     onBlur: onBlurCallback,
     onFocus: onFocusCallback,
@@ -98,7 +96,8 @@ const DefaultRange = ({
   }
 
   const rendererResponse = renderer({
-    innerProps: {...rangeProps, datalist}
+    id: rangeProps.id,
+    innerProps: rangeProps
   })
 
   // render custom component
@@ -109,8 +108,12 @@ const DefaultRange = ({
     <div
       className={`sui-FormBuilder-field sui-FormBuilder-DefaultRange sui-FormBuilder-${rangeProps.id}`}
     >
-      <MoleculeInputField {...fromInputProps} {...rendererResponse} />
-      <MoleculeInputField {...toInputProps} {...rendererResponse} />
+      <div className="sui-FormBuilder-DefaultRange-from">
+        <MoleculeInputField {...fromInputProps} {...rendererResponse} />
+      </div>
+      <div className="sui-FormBuilder-DefaultRange-to">
+        <MoleculeInputField {...toInputProps} {...rendererResponse} />
+      </div>
     </div>
   )
 }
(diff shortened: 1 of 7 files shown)


components/form/builder/src/Range/Default/index.js
```diff
@@ -0,0 +1,130 @@
...
+
+import PropTypes from 'prop-types'
+import {field, createComponentMemo} from '../../prop-types/index.js'
+
+import MoleculeInputField from '@s-ui/react-molecule-input-field'
+
+const joinInputValues = inputValues =>
+  inputValues ? Object.values(inputValues).join(',') : ''
+const splitInputValues = inputValues =>
+  inputValues?.split(',').reduce((acc, curr, index) => {
```
andresz1, line 11: `const [from, to] = inputValues?.split(',')`

components/form/builder/src/Range/Default/index.js
```diff
@@ -0,0 +1,132 @@
...
+import PropTypes from 'prop-types'
+import {field, createComponentMemo} from '../../prop-types/index.js'
+
+import MoleculeInputField from '@s-ui/react-molecule-input-field'
+
+const joinInputValues = inputValues =>
+  inputValues ? Object.values(inputValues).join(',') : ''
+
+const splitInputValues = inputValues => {
+  const [from, to] = inputValues?.split(',') || []
```
andresz1, line 12: If you set `[]` as a default param you won't need the `?` and the `|| []` since `[].split(',') = []`

Sabri1209: Yes, you're right, it's better that way!

components/form/builder/src/Range/Default/index.js
```diff
@@ -0,0 +1,132 @@
...
+    (ev, {value, name}) => {
+      if (name === fromInputId) {
+        setInputValues(prev => ({...prev, from: value}))
+      }
+      if (name === toInputId) {
+        setInputValues(prev => ({...prev, to: value}))
+      }
+    },
+    [fromInputId, toInputId]
+  )
```
andresz1, line 47: Is this a controlled or an uncontrolled component ? 🤔

components/form/builder/src/Range/Default/index.js
```diff
@@ -0,0 +1,127 @@
...
+      hidden: true
+    })
+  }
+
+  const fromInputProps = {
+    id: fromInputId,
+    name: fromInputId,
+    label: range.label || '',
+    value: inputValues?.from,
+    placeholder: 'Desde',
```
jmanrumartinez, line 66: Maybe it's interesting to use the placeholder from the field defined in the JSON instead of having it hardcoded here.

jmanrumartinez, line 66: Maybe it's interesting to use the placeholder from the field defined in the JSON instead of having it hardcoded here.

components/form/builder/src/Range/Default/index.js
```diff
@@ -0,0 +1,127 @@
...
+    onFocus: onFocusCallback,
+    helpText: range.help,
+    tabIndex
+  }
+
+  const toInputProps = {
+    id: toInputId,
+    name: toInputId,
+    value: inputValues?.to,
+    placeholder: 'Hasta',
```
jmanrumartinez, line 79: Same here.

jmanrumartinez, line 79: Same here.

components/form/builder/src/Range/Default/index.js
Sabri1209: The problem is that it's only one field

jmanrumartinez (approved):

stivaliserna (approved):

javiauso (approved):

components/form/builder/src/Range/Default/index.js
tomasmax: But you can add new config params to the JSON to be able to pass these two placeholders' values. It could be an improvement for the next iteration

tomasmax (approved):

PR #498, opened 2022-06-03
Use matchMedia as default value when is supported

giodelabarrera:
# Use `window.matchMedia` as default value when is supported.

## Context

The current behaviour of hook is that we always initialize [the state value `match`](https://github.com/SUI-Components/adevinta-spain-components/blob/master/components/react/hooks/src/useMediaQuery/index.js#L10) with [`defaultMatches` value](https://github.com/SUI-Components/adevinta-spain-components/blob/master/components/react/hooks/src/useMediaQuery/index.js#L6) and really we use `window.match` after render of the DOM by `useEffect`. That causes in client where `window.match` is supported a change of `match` value and renders of the two cases that are using the `match` value.

Example:
As you can see below in the images, console section, `Mobile` and `Desktop` components will be rendered.

Captura de pantalla 2022-04-07 a las 13 42 24
Captura de pantalla 2022-04-07 a las 13 41 27

Demo of example: https://codesandbox.io/s/great-forest-zd7syl?file=/src/App.js 

## Implementations:

- Use `window.match` when is supported as priorized value over defaultMatches.
- Move `supportMatchMedia` inside of hook for not to use its value by closure if the file `useMediaQuery/index.js` had already been loaded.
- Add unit test for `useMediaQuery` hook.
- Install [`css-mediaquery`](https://github.com/ericf/css-mediaquery) as dev dependency for support implementation of calculate of matches via screen with.

### DIFF
--- a/components/react/hooks/src/useMediaQuery/index.js
+++ b/components/react/hooks/src/useMediaQuery/index.js
@@ -1,13 +1,16 @@
 import {useState, useEffect} from 'react'
 
-const supportMatchMedia =
-  typeof window !== 'undefined' && typeof window.matchMedia !== 'undefined'
-
 function useMediaQuery(queryInput, {defaultMatches = false} = {}) {
-  const query = queryInput.replace(/^@media( ?)/m, '')
+  const supportMatchMedia =
+    typeof window !== 'undefined' && typeof window.matchMedia !== 'undefined'
   const matchMedia = supportMatchMedia ? window.matchMedia : null
 
-  const [match, setMatch] = useState(defaultMatches)
+  const query = queryInput.replace(/^@media( ?)/m, '')
+
+  const [match, setMatch] = useState(() => {
+    if (supportMatchMedia) return matchMedia(query).matches
+    return defaultMatches
+  })
 
   useEffect(() => {
     let active = true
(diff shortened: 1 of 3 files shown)


components/react/hooks/src/useMediaQuery/index.js
```diff
@@ -1,13 +1,16 @@
 import {useState, useEffect} from 'react'
 
-const supportMatchMedia =
```
midudev, line 3: No termino de entender por qué meter la constante dentro del hook. ¿Es un check que se quiere hacer cada vez que cambia el estado y por cada hook? ¿Da alguna ventaja?

components/react/hooks/src/useMediaQuery/index.js
```diff
@@ -1,13 +1,16 @@
...
-  typeof window !== 'undefined' && typeof window.matchMedia !== 'undefined'
-
 function useMediaQuery(queryInput, {defaultMatches = false} = {}) {
-  const query = queryInput.replace(/^@media( ?)/m, '')
+  const supportMatchMedia =
```
andresin87, line 4:
```suggestion
  const hasMatchMedia =
```
boolean const must include `has` or `is` prefixes

components/react/hooks/src/useMediaQuery/index.js
giodelabarrera: La ventaja que le veo es consistencia, ya que siempre aunque lo ejecute en cada render del hook ve en real time el soporte y no por closure de la primera ejecución.

PR #501, opened 2022-06-21
feat(components/form/builder): requestInterceptor returns a callback …

jmanrumartinez:
### Types of changes

- [ ] 🪲 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### Description, Motivation and Context
Our main motivation was to use a use case calling instead of making the HTTP request using the in-built fetch of formbuilder, to take advantage of object transformations, cache... 

carbon (1)

But the form builder doesn't support that, just a config object that it's used in the fetch method, so in order to not make a breaking change, we decided to provide a new param for the `requestInterceptor` called `callback` which will be resolved by the FB if it exists.

### DIFF
--- a/components/form/builder/src/reducer/rules.js
+++ b/components/form/builder/src/reducer/rules.js
@@ -24,7 +24,7 @@ const fetch = config =>
   new Promise((resolve, reject) => {
     const {url, headers, withCredentials = true} = config
     const request = new window.XMLHttpRequest()
-    request.onreadystatechange = function () {
+    request.onreadystatechange = function() {
       if (request.readyState === window.XMLHttpRequest.DONE) {
         if (request.status === 200) {
           resolve(request.response)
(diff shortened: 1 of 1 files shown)


components/form/builder/src/reducer/rules.js
```diff
@@ -118,60 +118,74 @@ export const shouldApplyRule = (fields, changeField, locale) => when => {
...
+        const config = requestInterceptorConfig || defaultConfig
+
+        const {remote, ...resetValue} = nextValue
+
+        if (callback) {
+          const response = await callback()
+
+          const nextJSON = await responseInterceptor({
+            field,
+            url: '',
```
jmanrumartinez, line 160: I leave the empty string in case it's expected to always be the `url` param on the `responseInterceptor` call, in order to not make any breaking change. 

In MA, for example, we check the incoming URL if includes some path to perform some actions.

quinwacca (approved):

javiauso (approved):

PR #516, opened 2022-09-01
Feat/scmi 75314/hidden links in link basic

nachoAdevinta:
Add prop to support hide link to crawlers in LinkBasic component.

The technique used is to use semantic `button` instead `a`. Documentation can be found here: 
[https://www.searchenginejournal.com/prg-pattern-the-new-nofollow/392862/#close](https://www.searchenginejournal.com/prg-pattern-the-new-nofollow/392862/#close)

image

### DIFF
--- a/components/link/basic/src/index.js
+++ b/components/link/basic/src/index.js
@@ -93,6 +93,7 @@ LinkBasic.propTypes = {
   className: PropTypes.string,
   disabled: PropTypes.bool,
   handleClick: PropTypes.func,
+  hideSemanticLink: PropTypes.bool,
   icon: PropTypes.element,
   literal: PropTypes.string,
   rel: PropTypes.string,
--- a/components/link/basic/src/index.js
+++ b/components/link/basic/src/index.js
@@ -2,6 +2,7 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 
 import {Link} from '@s-ui/react-router'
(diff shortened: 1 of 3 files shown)


andresin87:
This component is called Link-Basic. So if you need to change between button or link maybe a polymorphism paradigm will be a better option.

`@s-ui/react-primitive-polymorphic-element`

`as: PropTypes.elementType`


`const {as: As} = props`

`return (<Poly as={As}/>`

components/link/basic/src/index.js
```diff
@@ -51,6 +55,24 @@ export default function LinkBasic({
     )
   }
 
+  if (hideSemanticLink) {
+    const handleHiddenLinkClick = () => {
+      handleClick && handleClick()
```
andresin87, line 60: event must be provided (agreements)

components/link/basic/src/index.js
```diff
@@ -25,49 +27,64 @@ export default function LinkBasic({
   useReactRouterLinks,
   url = '#'
 }) {
-  const linkClassName = cx('sui-LinkBasic', className)
+  const BASE_CLASS = 'sui-LinkBasic'
+  const linkBasicClassName = cx(BASE_CLASS, className)
```
andresin87, line 31: `@s-ui/component-dependencies`is no longer including classnames package. Plz add the 2.2.5 version to the dependencies.

andresin87 (approved):

oriolpuig (approved):

PR #521, opened 2022-09-28
feat(components/form/builder): …

Sabri1209:
Merges field props with the rest before exposing it through the renderer to be able to handle all fields in an isolated way


components/form/builder/src/Standard/Fields/Multifield/index.js
```diff
@@ -0,0 +1,43 @@
...
+  errors,
+  alerts,
+  renderer
+}) => {
+  const {fields} = field || {}
+
+  const rendererResponse = renderer({
+    id: field.id,
+    type: field.type,
+    innerProps: fields.map(props => ({
```
javiauso, line 18: IMHO should be called with ProxyField so that each field can determine its own representation

Sabri1209: I like it! Thanks @javiauso

PR #528, opened 2023-01-03
fix(components/form/builder): pickers autocomplete/dropdown are shown…

oriolpuig:
## Form/Builder 
`❓ Ask`

**TASK**: NO TASK ID

### Types of changes

- [x] 🪲 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [x] 📷 Demo
- [x] 🧪 Test

#### Context

As you can see at the image attached, the `Marca` field has 2 error messages "Debes seleccionar una marca". ~(⚠️ Spoiler: That's the 🐛)~ . This field is a `type=Picker display=Dropdown` field. But wait, if the `Modelo` field is also a `type=Picker display=Dropdown`, why has not 2 errors also? The answer is.... because it is disabled!

Beeing more concise, [at those lines](https://github.com/SUI-Components/adevinta-spain-components/blob/master/components/form/builder/src/Standard/index.js#L61-L77), our `checkConstraintsFromField` method is validating our DOM Element and seeing the native [ValidityState](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState). Those lines will tell us if the DOM Element has errors such as `badInput`, `valueMissing` and so on. 

Continuing [with those other lines](https://github.com/SUI-Components/adevinta-spain-components/blob/master/components/form/builder/src/Standard/index.js#L98-L115), here we have a custom edge validation for `type=Picker display=Dropdown` fields. If we read the comment line... we will understand why we need this: `// custom validation: since SUI select and autocomplete add disabled and readOnly props to input the default validity function marks allways as valid the field` 

At some point in 2022, the internal structure of MoleculeSelect had a input field with disabled / readonly attribute causing us to create a manual validation at `form/builder`. That structure has been changed/improved and now the initial validation (the native one) is checking well the validity of this `type=Picker display=Dropdown` fields ; and currently we don't need the second validation. 

And that's why the `Marca` field has 2 errors (the native one and our custom edge case for pickers) and the `Modelo` only has one, because was disabled. 
 
#### Proposal
I propose to remove our custom edge validation in favour of native one. I thought also to clean the array to prevent duplicated error messages but I think this is the better and clean solution.

### Screenshots - Animations

Image of Marca field with two error messages and Modelo field only with one

### DIFF
--- a/components/form/builder/mocks/index.js
+++ b/components/form/builder/mocks/index.js
@@ -1,0 +1,53 @@
+export const FORM_BUILDER_SELECT_FIELD_MOCK = {
+  form: {
+    id: 'milanuncios-insert-car-v3',
+    type: 'group',
+    label: 'Publica tu anuncio',
+    actionlabel: 'Publicar',
+    fields: [
+      {
+        id: 'hasEnergyCertificate',
+        hint: '¿Tienes certificado energético?',
+        type: 'picker',
+        tabIndex: 0,
+        display: 'dropdown',
+        required: true,
+        datalist: [
+          {
+            value: 'true',
+            text: 'Sí, tengo el certificado energético'
+          },
+          {value: 'false', text: 'Está en trámite'}
+        ],
+        constraints: [
+          {
+            property: {notnull: ''},
+            message: 'Este campo es obligatorio'
+          }
+        ]
+      },
+      {
+        id: 'searchACountry',
+        hint: 'Busca un país',
+        type: 'picker',
+        tabIndex: 0,
+        display: 'autocomplete',
+        required: true,
+        datalist: [
+          {
+            value: 'spain',
+            text: 'España'
+          },
+          {value: 'france', text: 'Francia'}
+        ],
+        constraints: [
+          {
+            property: {notnull: ''},
+            message: 'Este campo es obligatorio'
+          }
+        ]
+      }
+    ],
+    rules: {}
+  }
+}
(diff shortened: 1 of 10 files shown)


aleixferr (approved):

components/form/builder/mocks/index.js
```diff
@@ -0,0 +1,108 @@
+export const FORM_BUILDER_SELECT_FIELD_MOCK = {
```
javiauso, line 1: Maybe this file could fit better in `demo` dir instead `src`

javiauso (approved):

stivaliserna (approved):

components/form/builder/test/index.test.js
```diff
@@ -1,22 +1,93 @@
...
+      const {queryAllByText, getByText} = setupForm(props)
+
+      const button = getByText(/Submit/)
+      await userEvent.click(button)
+      const requiredFields = queryAllByText('Este campo es obligatorio')
+
+      expect(requiredFields).to.have.lengthOf(2)
+    })
+
+    it('when we aa required field is disabled', async () => {
```
stivaliserna, line 75: Maybe `Should not display error validation message when field is disabled` (?)

stivaliserna:
Great job adding tests and improving the demo <3

components/form/builder/mocks/index.js
oriolpuig: Hi @javiauso, thanks for your reply 🙏🏻 

This mock files are not inside `src` file. This is inside new `mocks` folder and ignored from `npm` file to avoid increase npm package. The main reason of that is to share mocks between `demo` and `tests` contexts instead of have it inside `demo` folder. 

That's why I create this new contextual folder.

javiauso: Ups you're right!! 😆

PR #529, opened 2023-01-18
feat(components/login/passwordReset): Create scaffolding for passwordReset component

oegea:
From CNET PRO, we are working to implement a generic password reset component that will fully render the interface to execute the password recovery flow.

This PR starts creating the scaffolding.


components/login/passwordReset/README.md
```diff
@@ -0,0 +1,36 @@
+# LoginPasswordReset
+
+Work in progress.
```
jordevo, line 3: 🍆 something we could already add at this point, right?

davidmartin84 (approved):

components/login/passwordReset/README.md
oegea: We'll add ASAP 😅 
As we'll be working on this on the following days, a lot of Prs will be submitted to add features and complete the component. This is just a fast scaffolding to ensure we share the same component structure.

components/login/passwordReset/src/index.scss
```diff
@@ -0,0 +1,5 @@
+@import '~@s-ui/theme/lib/index';
+
+.sui-LoginPasswordReset {
```
jordevo, line 3:
```suggestion
.mt-LoginPasswordReset {
```

🍆  should be like this, right?

jordevo (approved):

components/login/passwordReset/src/index.scss
davidmartin84: This component is intended to be global. reusable for all verticals. Maybe we can think if the prefix "sui-" is the best... but coupling it to "mt" is not the idea.

jordevo: you're absolutely right! I missed we're on this repo! 😅

PR #541, opened 2023-01-23
feat(components/login/passwordReset): add hook to check info stage in…

davidmartin84:
…fo step

### DIFF
--- a/components/login/passwordReset/src/hooks/components/useGetStageInfoStep.js
+++ b/components/login/passwordReset/src/hooks/components/useGetStageInfoStep.js
@@ -1,0 +1,52 @@
+import {STATUSES} from '@s-ui/react-molecule-progress-steps'
+
+import {
+  STAGE_PASSWORD_CHANGE,
+  STAGE_PASSWORD_RESET_START
+} from '../../config.js'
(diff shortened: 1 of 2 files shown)


components/login/passwordReset/src/hooks/components/useGetStageInfoStep.js
```diff
@@ -0,0 +1,52 @@
...
+        stage === STAGE_PASSWORD_RESET_START
+          ? STATUSES.ACTIVE
+          : STATUSES.VISITED
+    }
+
+    const endStepperState = {
+      label: i18n.t('LOGIN_CROSS.PASSWORD_RESET.STEP_2.LABEL'),
+      status:
+        stage === STAGE_PASSWORD_CHANGE ? STATUSES.ACTIVE : STATUSES.NORMAL
+    }
```
oegea, line 37: IMHO, it could be cleaner to have one single condition to determine which is the current stage, and then create an object with title, message, initialStepperState and endStepperState properties for each kind of state.

Something like this...

```
if (stage === 1) {
   data = {
     title: '1',
     message: '1'
   }
} else {
   data = { title: '2', message: '2' }
}

return data
```

oegea (approved):

PR #544, opened 2023-01-31
feat(components/login/header): New component: login header

davidmartin84:
Captura de Pantalla 2023-01-31 a las 13 43 46


components/login/header/test/index.test.js
```diff
@@ -0,0 +1,56 @@
+/*
```
davidmartin84, line 1: WIP: Tests will be done in another task

oegea (approved):

components/login/header/README.md
```diff
@@ -0,0 +1,34 @@
+# LoginHeader
```
jordevo, line 1: should we keep this component on an already existing context, such as `topbar` or `header`? what do you think?

davidmartin84: This header is only for the "login" context. In the future, the cross team can decide whether to place it in another context or in its own repository.

PR #569, opened 2023-02-13
Feat/mmaa 33148 add signed url prop to map google

lolarufino:
In order to use a static google map in pages where we have many petitions, it is recommended a digital signature - in addition to an API key - to authenticate requests. 
So, with this PR we add a new prop to this component. `signedUrl` prop. 
This prop has priority over the rest of the props that generate the url when you don't need a signature. 
Pay attention to the [digital signature documentation](https://developers.google.com/maps/documentation/maps-static/digital-signature?hl=en) to see how to create this url in your web app.


ivanmlaborda (approved):

fredyzz (approved):

oriolpuig (approved):

andresin87:
All google maps features should be applied on 
https://github.mpi-internal.com/scmspain/frontend-common--uilib-components

PR #589, opened 2023-05-18
test(components/tool/taskManager): Add tests for run simple task use case

oegea:
# Context

Now that we have agreed that TaskManager is the kind of solution we want to adopt on a mid-term, we are working to make the component more stable and usable.

These are a set of PRs to ensure the task manager system has a well-enough test base.

# Changes

This PR adds a test to the `run simple task` use case.


jordevo (approved):
glad to see some component & domain testing added! 👏 

I wonder when the component domain test spec is supposed to run 🤔 is it somehow run on CI?

emiliovillu (approved):

PR #592, opened 2023-05-19
Test/add taskmanager tests

oegea:
# Context

Now that we have agreed that TaskManager is the kind of solution we want to adopt on a mid-term, we are working to make the component more stable and usable.

These are a set of PRs to ensure the task manager system has a well-enough test base.

# Changes

This PR adds a test to CancelWorkTaskUseCase and FinishWorkTaskUseCase.


quinwacca (approved):

components/tool/taskManager/src/domain/test/task/FinishWorkTaskUseCaseSpec.js
```diff
@@ -0,0 +1,66 @@
...
+    const stateResult = await finishWorkTaskUseCase.execute({
+      localState,
+      workId: localState.tasks[0].work[0].id,
+      taskId: localState.tasks[0].id
+    })
+
+    // Then
+    const task = stateResult.tasks[0]
+    const work = task.work[0]
+    expect(work.status).to.eql(config.get('AVAILABLE_STATUS').COMPLETED)
```
jordevo, line 37: 🍆 I might be a bit late on this but perhaps `job` would be a better name for what this is supposed to mean 😅

jordevo (approved):

PR #611, opened 2023-09-04
feat(components/tool/taskManager): update layout styles taskManager

davidmartin84:
**WIP!!!!**


components/tool/taskManager/src/domain/config.js
```diff
@@ -8,9 +10,9 @@ export default class Config {
     this._config = {
       AVAILABLE_STATUS: {
         CANCELLED: 'CANCELLED',
-        COMPLETED: 'COMPLETED 🥳 ',
+        COMPLETED: <ChevronCircleDownSolid size="medium" />,
```
oegea, line 13: 🍆 This is a bit weird, as domain config should be decoupled from react things. However, could be a tradeoff maybe...

davidmartin84: agree! in next commit i will refactor this

PR #620, opened 2023-09-21
feat(components/tool/taskManager): Add the possibility to filter task…

oegea:
# Changes

Add the possibility of filtering tasks to display only certain number of tasks, or filter them by id.

### DIFF
--- a/components/tool/taskManager/src/index.js
+++ b/components/tool/taskManager/src/index.js
@@ -12,7 +12,12 @@ import useBeforeUnloadEffect from './hooks/useBeforeUnloadEffect.js'
 import useContext from './hooks/useContext.js'
 import useDevMode from './hooks/useDevMode.js'
 
-export default function ToolTaskManager({isVisible = true, statusIcons = {}}) {
+export default function ToolTaskManager({
+  isVisible = true,
+  statusIcons = {},
+  taskIdsFilter = [],
+  taskAmountFilter = 0
+}) {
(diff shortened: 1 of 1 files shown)


davidmartin84 (approved):

jordevo (approved):

components/tool/taskManager/src/index.js
```diff
@@ -12,16 +12,31 @@ import useBeforeUnloadEffect from './hooks/useBeforeUnloadEffect.js'
...
     return (
       <>
-        {state.tasks.map(task => {
+        {state.tasks.map((task, index) => {
           if (task.visibleWork === 0 && isDevModeEnabled === false) return
+
+          if (
+            Array.isArray(taskIdsFilter) &&
+            taskIdsFilter.length > 0 &&
+            taskIdsFilter.find(id => id === task.id) === undefined
```
jordevo, line 34:
```suggestion
            Array.isArray(taskIdsFilter) &&
            taskIdsFilter.some(id => id === task.id)
```

would this work for you and perhaps simplify the logic? 🤔

jordevo (approved):

components/tool/taskManager/src/index.js
oegea: At first we had the statement that way, but we prefered to not filter tasks at all if the array was empty, to keep things simple by default.

PR #623, opened 2023-09-28
feat(components/tool/taskManager): Avoid displaying alerts when task …

oegea:
When displaying a message to indicate that all tasks have been finished, ignore not-visible tasks.

### DIFF
--- a/components/tool/taskManager/src/hooks/useState.js
+++ b/components/tool/taskManager/src/hooks/useState.js
@@ -61,7 +61,10 @@ const useState = ({onCompleteAllTasks}) => {
         next(result)
       })
 
-  const checkIfAllTasksHaveBeenFinished = state => {
+  const checkIfAllTasksHaveBeenFinished = (taskId, state) => {
+    // No notice on not-visible tasks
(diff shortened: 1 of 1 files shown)


davidmartin84 (approved):

jordevo (approved):

components/tool/taskManager/src/hooks/useState.js
```diff
@@ -61,7 +61,10 @@ const useState = ({onCompleteAllTasks}) => {
         next(result)
       })
 
-  const checkIfAllTasksHaveBeenFinished = state => {
+  const checkIfAllTasksHaveBeenFinished = (taskId, state) => {
+    // No notice on not-visible tasks
```
jordevo, line 65: not sure what does this mean... _no notification_? 🤔

oegea: Yep, probably using a bad word.

oegea: This means that callback won't be executed

PR #675, opened 2024-11-21
feat(components/form/builder): Fix Form Builder init state when React 18

javiauso:
Fix Form Builder state when **React 18**.

With React 18 new **batch rendering**, when the component set the state fields with initial fields at mount, **only update the last field** because of the process read a ref instead the updated state in each iteration.

Before it, in some cases, you could find that a form works well, but it's happening when is applying rules at the first render.

**BEFORE:**

Screenshot 2024-11-21 at 09 26 14

**AFTER:**

Screenshot 2024-11-21 at 09 26 47

### DIFF
--- a/components/form/builder/src/index.js
+++ b/components/form/builder/src/index.js
@@ -1,22 +1,22 @@
-import {useCallback, useEffect, useRef, useState} from 'react'
+import {useCallback, useEffect, useState} from 'react'
 
 import PropTypes from 'prop-types'
 
 import {inputSizes as fieldSizes} from '@s-ui/react-atom-input'
 import AtomSpinner, {atomSpinnerTypes} from '@s-ui/react-atom-spinner'
 
-import {getUpdatedFormState} from './mapper/formState'
-import {RULES} from './reducer/constants'
+import {getUpdatedFormState} from './mapper/formState.js'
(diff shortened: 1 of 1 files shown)


tomasmax (approved):

nucliweb (approved):

components/form/builder/src/index.js
```diff
@@ -45,11 +45,9 @@ const FormBuilder = ({
   const {fields = [], rules = {}, id: formID} = json.form
   const [stateFields, setStateFields] = useState(fields)
   const [stateShowSpinner, setStateShowSpinner] = useState(false)
-  const __PERFORMANCE_UGLY_HACK_STATE_FIELDS__ = useRef()
-  __PERFORMANCE_UGLY_HACK_STATE_FIELDS__.current = stateFields
 
-  const changeField = (fieldId, fieldValue) => {
-    const nextFields = changeFieldById(__PERFORMANCE_UGLY_HACK_STATE_FIELDS__.current, fieldId, {value: fieldValue})
```
kikoruiz, line 52: 😮

kikoruiz (approved):

lolarufino (approved):

PR #691, opened 2025-05-08
feat(components/form/builder): errorText and alerText will be empty s…

quinwacca:
## Form / Builder

With a previous PR, the empty value of `errorText` and `alertText` of every field was modified from the expected **empty string** to **empty array**. 
And it is painting fields like error fields, so I modified again to return the expected empty value ( `''` ).

image


tomasmax (approved):

andresin87 (commented):
✅

PR #697, opened 2025-05-26
feat(components/topbar/user): adding heading to topbar

ivanmlaborda:
A11Y
Add prop to conditionally render a heading element on UserTopbar component

https://jira.ets.mpi-internal.com/browse/MTR-96865

### DIFF
--- a/components/topbar/user/src/index.js
+++ b/components/topbar/user/src/index.js
@@ -24,8 +24,19 @@ const BODY_HAS_SCROLL_DISABLED = 'body-has-scroll-disabled'
 const TITLE_CLASS_NAME = 'sui-TopbarUser-title'
 const FLOW_BUTTON_CLASS_NAME = 'sui-TopbarUser-navButton'
 
+const HEADING_TAGS = {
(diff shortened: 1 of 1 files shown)


emiliovillu (approved):

components/topbar/user/src/index.js
```diff
@@ -24,6 +24,33 @@ const BODY_HAS_SCROLL_DISABLED = 'body-has-scroll-disabled'
...
+    <Link href={url} className="sui-TopbarUser-brandIcon" title={name}>
+      <BrandIcon />
+    </Link>
+  ) : (
+    <Link href={url} className="sui-TopbarUser-brand" title={name}>
+      {BrandImage ? <BrandImage /> : name}
+    </Link>
+  )
+
+  return asHeading ? <Heading className="sui-TopbarUser-heading">{brandLink}</Heading> : brandLink
```
andresin87, line 51: This can affect the default margins of the brand when wrapped into each tag. Is it considered?
(of course it is retro-compatible, I know, just for understanding)

PR #719, opened 2026-04-23
feat(components/topbar/user): add navCTASecondary prop

stivaliserna:
This PR:
- adds a `navCTASecondary` prop (node) to `TopbarUser`, rendered in the cta area after `navCTA`.
- Hidden when `customContent` is provided.
- Aligns vertically cta buttons.

### DIFF
--- a/components/topbar/user/src/index.js
+++ b/components/topbar/user/src/index.js
@@ -87,6 +87,7 @@ export default function TopbarUser({
   navButton,
   navCTA,
   navCTALogin,
+  navCTASecondary,
   navMain,
   navUser,
   onToggle = () => {},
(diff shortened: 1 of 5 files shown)


components/topbar/user/src/index.js
```diff
@@ -307,6 +308,9 @@ export default function TopbarUser({
             </AtomButton>
           </div>
         )}
+        {navCTASecondary && !customContent && (
+          <>{navCTASecondary ? <div className="sui-TopbarUser-ctaSecondary">{navCTASecondary}</div> : null}</>
```
maiderhernandorena-del: Do we need 2nd validation here? isn't enough like this:
````
{navCTASecondary && !customContent && (
   <div className="sui-TopbarUser-ctaSecondary">{navCTASecondary}</div>
)}
````

maiderhernandorena-del (changes requested):

maiderhernandorena-del (approved):

